PR #28, opened 2017-03-10
SignTool: Use LocateVS to find msbuild when not specified

tmat:
Removes -dotnetPath command line arg, since the microbuild real signing task won't be portable anytime soon.


tmat:
@jaredpar @tannergooding

tmat:
Never mind, needs a bit more work...

PR #41, opened 2017-07-03
Add rebuild support to repo toolset

333fred:
Allows users to specify doing a full rebuild. This change is split into two commits, one for the gitignore and whitespace fixes, and one for the actual content change. Tagging @tmat for review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,7 @@
 *.lock.json
 *.nuget.props
 *.nuget.targets
-NuGet.exe
-nuget.exe
+[Nn]u[Gg]et.exe
 [Pp]ackages/
 
 ###############################################################################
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -14,6 +14,7 @@
 *.nuget.props
 *.nuget.targets
 NuGet.exe
+nuget.exe
```
tmat: nuget.exe shouldn't be used at all. The toolset uses msbuild or dotnet cli to restore. How does nuget show up?

333fred: https://github.com/dotnet/roslyn-tools/blob/master/build/Scripts/Windows/Build.ps1#L153. Really, I should change this to be [Nn]u[Gg]et.exe

333fred: Because I use bash, and git for linux treats NuGet.exe and nuget.exe as different files.

tmat: Oh, I see. This repo itself doesn't use RepoToolset (it should!). Looks good.

PR #42, opened 2017-07-05
Add "skipIfEqualOrNewerInstalled" command line switch...

KevinH-MS:
...and return unique exit codes for each type of failure.

### DIFF
--- a/src/VSIXExpInstaller/VSIXExpInstaller/Program.cs
+++ b/src/VSIXExpInstaller/VSIXExpInstaller/Program.cs
@@ -237,7 +237,7 @@ namespace VsixExpInstaller
                                 var installableExtension = extensionManager.CreateInstallableExtension(extensionPath);
 
                                 var status = GetInstallStatus(installableExtension.Header);
-                                var installedVersionIsEqualOrNewer = installableExtension.Header.Version < status.installedExtension.Header.Version;
+                                var installedVersionIsEqualOrNewer = installableExtension.Header.Version < status.installedExtension?.Header.Version;
 
                                 if (uninstall)
                                 {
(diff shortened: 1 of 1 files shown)


KevinH-MS:
FYI @tannergooding 
/CC @dotnet/roslyn-infrastructure 

Background:  For the LUT repo, I want to be able to call VsixExpInstaller to deploy a matching or newer version of Roslyn than what I build against, but short-circuit (with a message) if the version installed is equal or newer than what I already have.

KevinH-MS:
FYI @tannergooding I've updated this to also do the "up to date" check on uninstall.  Please take a look.

KevinH-MS:
Ping for review...

src/VSIXExpInstaller/VSIXExpInstaller/Program.cs
```diff
@@ -49,14 +57,15 @@ private static bool FindArg(List<string> args, string argName)
...
-            Console.WriteLine("Usage: VSIXExpInstaller [/rootSuffix:suffix] [/u] [/uninstallAll] [/vsInstallDir:vsInstallDir] <vsix>");
+            Console.WriteLine("Usage: VSIXExpInstaller [/rootSuffix:suffix] [/skipIfEqualOrNewerInstalled] [/u] [/uninstallAll] [/vsInstallDir:vsInstallDir] <vsix>");
             Console.WriteLine("Suffix is Exp by default.");
         }
 
         private static string GetDevenvPath(string vsInstallDir)
         {
             if (string.IsNullOrWhiteSpace(vsInstallDir))
             {
+                Environment.ExitCode = E_ACCESSDENIED;
```
tannergooding: nit: I thought you could set the ErrorCode of an exception and it would do the same...

KevinH-MS: Didn't know that.  I will try.

KevinH-MS: It doesn't *seem* to work that way.  I tried throwing a Win32Exception and set ErrorCode, but the ERRORLEVEL was still -532462766.

PR #50, opened 2017-08-31
ModifyVsixManifest: Update hash in manifest.json

tmat:
Update SHA256 hash of extension.vsixmanifest file that is stored in manifest.json of the VSIX being modified.


tmat:
@tannergooding @jmarolf Please review

PR #55, opened 2017-09-07
enable F# support

brettfo:
Enable F# support.

@tmat @KevinRansom


src/SignTool/SignTool/Program.cs
```diff
@@ -200,7 +200,8 @@ internal static void PrintUsage()
             // Get defaults for all of the optional values that weren't specified
             if (msbuildPath == null && !test)
             {
-                var vsInstallDir = LocateVS.Instance.GetInstallPath("15.0", new[] { "Microsoft.Component.MSBuild" });
+                // installed VS versions could be "15.0" or "15.3", so "15." is a good proxy
```
tmat: This one isn't necessary. The toolset finds the VS installation using vswhere tool and passes it to SignTool explicitly.

tmat:
👍 Looks great. Simpler than I expected!

src/SignTool/SignTool/Program.cs
brettfo: With only VS 15.3 installed `build.cmd` fails locally because only VS with versions `"15.3"` could be found which doesn't match the `.StartsWith()` check [here](https://github.com/dotnet/roslyn-tools/blob/27a7b5e0fa8e0c625435c7f07f85efcb99f65f24/src/LocateVS/LocateVS/Instance.cs#L56).

Versions 15.1 and 15.2 of VS still reported `"15.0"` as the product version.

tmat: That's odd, since the msbuildPath should be set and LocateVS shouldn't be called at all.

PR #60, opened 2017-10-18
Updating Repo Toolset to contain various fixes.

tannergooding:
FYI. @tmat, @jaredpar, @brettfo, @nguerrera 

Most of these issues were found when moving the SDK to repo toolset:
* Including imported props/targets as part of MSBuildAllProjects is best practice (CC @nguerrera)
* DOTNET_INSTALL_DIR is used by the SDK, other repos, and the dotnet-install.ps1 script and needs to be respected
* ProjectLanguage is used heavily by the SDK/CLI repositories
* The SDK won't add the three default NoWarn values (such as CS1701) if NoWarn is already set. This can cause various failures when compiling netcore projects
* Not being able to use the netcore toolset compiler and not being able to opt-out of using a toolset compiler can break some builds
* The XUnit beta version we were on had an xml syntax bug in one of its targets and broke certain builds.


src/RepoToolset/DefaultVersions.props
```diff
@@ -1,30 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tmat: Is this <xml> tag necessary? I'd rather avoid it.

src/RepoToolset/GenerateResxSource.targets
```diff
@@ -48,4 +54,5 @@
       </EmbeddedResourceSG>
     </ItemGroup>
   </Target>
+
 </Project>
```
tmat: Extra empty line

src/RepoToolset/DefaultVersions.props
```diff
@@ -1,30 +1,41 @@
...
 <!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
 <Project>
+
   <!--
     Sets versions of packages used by the toolset unless they were already specified by the repo.
   -->
+
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+  </PropertyGroup>
```
tmat: I'd rather not add this everywhere. The toolset files won't change in the repo unless the repo restores a new version of the package.

src/RepoToolset/ProjectDefaults.props
```diff
@@ -61,18 +68,19 @@
...
-  
+
   <Choose>
     <!-- C# specific settings -->
     <When Condition="'$(Language)' == 'C#'">
       <PropertyGroup>
         <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
-        <NoWarn>$(NoWarn);1591</NoWarn>
+        <NoWarn Condition="'$(NoWarn)' == ''">1701;1702;1705;1591</NoWarn>
+        <ProjectLanguage>CSharp</ProjectLanguage>
```
tmat: ><ProjectLanguage>CSharp</ProjectLanguage> [](start = 8, length = 41)

What is this property used for?

src/RepoToolset/Tools.props
```diff
@@ -1,9 +1,31 @@
...
-    <PackageReference Include="Microsoft.Net.Compilers" Version="$(MicrosoftNetCompilersVersion)" />
-  </ItemGroup>
+
   <PropertyGroup>
-    <ToolsetCompilerPackageDir>$(NuGetPackageRoot)microsoft.net.compilers\$(MicrosoftNetCompilersVersion)\</ToolsetCompilerPackageDir>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
+
+  <Choose>
+    <When Condition="'$(UsingToolMicrosoftNetCompilers)' != 'netcore'">
```
nguerrera: The values for this are true, false, netcore? That's confusing.

src/RepoToolset/Tools.props
```diff
@@ -1,9 +1,31 @@
...
+      <ItemGroup>
+        <PackageReference Include="Microsoft.Net.Compilers.netcore" Version="$(MicrosoftNetCompilersNetCoreVersion)" />
```
nguerrera: Why do I get this package ref if the property is *not* netcore?

src/RepoToolset/RepoLayout.props
```diff
@@ -1,11 +1,22 @@
...
+
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- Respect environment variable for the .NET install directory if set; otherwise, use the current default location -->
+    <DotNetRoot Condition="'$(DotNetRoot)' == ''">$(DOTNET_INSTALL_DIR)</DotNetRoot>
+    <DotNetRoot Condition="'$(DotNetRoot)' == ''">$(RepoRoot).dotnet\</DotNetRoot>
+    <DotNetRoot Condition="!HasTrailingSlash('$(DotNetRoot)')">$(DotNetRoot)\</DotNetRoot>
```
tmat: I'd rather be more prescriptive here and force all repos to download dotnet to \.dotnet directory, unless the location is supplied by source build. Is source build gonna use the environment variable DOTNET_INSTALL_DIR?

src/RepoToolset/Settings.props
```diff
@@ -1,15 +1,22 @@
...
   <Import Project="DefaultVersions.props"/>
   <Import Project="ProjectLayout.props" />
   <Import Project="StrongName.props"/>
   <Import Project="Version.props"/>
   <Import Project="ProjectDefaults.props"/>
-  <Import Project="Tools.props" />
   <Import Project="Tests.props" />
-  
+
+  <Import Project="Tools.props" Condition="'$(UsingToolMicrosoftNetCompilers)' != 'false'" />
```
tmat: >Tools [](start = 19, length = 5)

Rename the file to Compilers.props, since it's now conditioned on UsingToolMicrosoftNetCompilers

src/RepoToolset/Tools.props
tmat: +1

When do we need to use netcore compilers? When building on Linux/MacOS? We can detect that condition automatically, instead of requiring the repo to chose manually.

---
In reply to: [145545980](https://github.com/dotnet/roslyn-tools/pull/60#discussion_r145545980) [](ancestors = 145545980)

src/RepoToolset/DefaultVersions.props
```diff
@@ -1,30 +1,41 @@
...
+
   <PropertyGroup>
     <!-- Opt-out features -->
     <UsingToolXliff Condition="'$(UsingToolXliff)' == ''">true</UsingToolXliff>
     <UsingToolSourceLink Condition="'$(UsingToolSourceLink)' == ''">true</UsingToolSourceLink>
     <UsingToolXUnit Condition="'$(UsingToolXUnit)' == ''">true</UsingToolXUnit>
+    <UsingToolMicrosoftNetCompilers Condition="'$(UsingToolMicrosoftNetCompilers)' == ''">true</UsingToolMicrosoftNetCompilers>
```
tmat: ><UsingToolMicrosoftNetCompilers Condition="'$(UsingToolMicrosoftNetCompilers)' == ''">true</UsingToolMicrosoftNetCompilers> [](start = 4, length = 123)

I think we should keep this as a non-optional feature. ALl repos must use the Roslyn compilers. We should auto-detect when NetCore package is needed vs. Desktop package.

src/RepoToolset/DefaultVersions.props
tannergooding: It lets editors that are not VS recognize the file as XML and behave correctly.

It is very painful to update these files otherwise.

src/RepoToolset/GenerateResxSource.targets
tannergooding: The convention in all the shipping props/targets has been to have an extra newline after the Project open tag and before the Project close tag.

src/RepoToolset/DefaultVersions.props
tmat: OK.

---
In reply to: [145550454](https://github.com/dotnet/roslyn-tools/pull/60#discussion_r145550454) [](ancestors = 145550454)

src/RepoToolset/DefaultVersions.props
tannergooding: This is best practice to have, regardless of whether the files should change.

src/RepoToolset/Tools.props
tannergooding: ah, typo

src/RepoToolset/GenerateResxSource.targets
tmat: That's an odd convention.

---
In reply to: [145550595](https://github.com/dotnet/roslyn-tools/pull/60#discussion_r145550595) [](ancestors = 145550595)

src/RepoToolset/DefaultVersions.props
tmat: I disagree. If there is no actual reason to have these then they are just clutter.

---
In reply to: [145551050](https://github.com/dotnet/roslyn-tools/pull/60#discussion_r145551050) [](ancestors = 145551050)

src/RepoToolset/RepoLayout.props
tannergooding: The environment variable is part of the official `dotnet-install.ps1` script and needs to be respected.

src/RepoToolset/DefaultVersions.props
tannergooding: All repos must use the Roslyn compilers. Not all repos must use the toolset compiler.

tmat: Why not? Other than Roslyn repo, where would that compiler come from?

---
In reply to: [145553642](https://github.com/dotnet/roslyn-tools/pull/60#discussion_r145553642) [](ancestors = 145553642)

src/RepoToolset/Tools.props
nguerrera: Right now if we run core msbuild, we can't build satellites with 2.0 SDK. (Fixed in 2.1+, but source-build won't be using that just yet.). So the condition would have to be MSBuildRuntimeType == Core, but then it would slow down builds that can use dotnet + desktop csc because core csc is quite a bit slower.

src/RepoToolset/RepoLayout.props
tmat: Will source build use that environment variable?

---
In reply to: [145553523](https://github.com/dotnet/roslyn-tools/pull/60#discussion_r145553523) [](ancestors = 145553523)

src/RepoToolset/DefaultVersions.props
nguerrera: Why must all repos use the package? If I'm getting a .NET Core SDK LKG, why can't I just use the compiler that comes with it. Why must I have two versions to maintain?

nguerrera: Another thing: Microsoft.Net.Compilers.netcore nupkg edition won't be crossgen'ed and it will slow down core builds even more. I should be allowed to use the crossgen'ed copy that's bundled with .NET Core SDK.

src/RepoToolset/Tools.props
tmat: I see, so what we need is an opt-in for core csc. I'd then define UsingToolNetCoreCompilers = true/false. 

---
In reply to: [145554104](https://github.com/dotnet/roslyn-tools/pull/60#discussion_r145554104) [](ancestors = 145554104)

src/RepoToolset/DefaultVersions.props
tmat: I see. Then do we ever need the NetCore package? Either repo can use desktop csc from Microsoft.Net.Compilers or it opts into using NetCore compilers. In the latter case we use the compilers bundled with .NET Core SDK.

src/RepoToolset/Tools.props
tannergooding: Do you have a better suggestion. You don't want to allow importing both, and the only other option that I can think of is a task to validate both `UsingToolMicrosoftNetCompilers` and `UsingToolMicrosoftNetCompilersNetCore` are not simultaneously true

src/RepoToolset/ProjectDefaults.props
tannergooding: SDK/CLI use it in a number of places. Someone who has been on the team longer may be able to comment if it is actually important.

tmat: Perhaps the repo should use ```$(Language)``` instead, or define ProjectLanguage in it's private props files if that makes things cleaner for some reason. I would rather avoid polluting all repos with yet another property that specifies the language.

src/RepoToolset/Tools.props
tmat: Yes, see my comment below. Add ```UsingToolNetCoreCompilers``` property. If true use the compiler that comes with dotnet SDK, otherwise use Microsoft.Net.Compilers package.

nguerrera: What about when building with desktop msbuild and UsingToolNetCoreCompilers=true. Is that an error case?

tmat: I'd think so.

src/RepoToolset/DefaultVersions.props
tannergooding: You may want to consume the bits in box, which are the same as the toolset compiler, but are ngen'd and which are faster.

You may also have a case (source build) where you need to be using a very explicit copy of the compiler.

src/RepoToolset/RepoLayout.props
tannergooding: I don't know, but like NUGET_PACKAGES. It is a well defined variable used by one of our shipping products, so we should respect it.

src/RepoToolset/DefaultVersions.props
tannergooding: It can also change if the user manually modifies the files on disk, which is the same case for the Common.targets/etc.

tannergooding: This can happen if a user is testing changes locally, for example.

tannergooding:
Updated. Turns out, `Microsoft.Net.Compilers.netcore` is not the netcore toolset compiler and it should never have been uploaded to nuget.

The actual netcore toolset compiler is `Microsoft.NetCore.Compilers` and it is not RTM ready yet.

The option is still `UsingToolMicrosoftNetCompilers`, and it only opts out of using the desktop toolset compiler. This is desirable, both when a user wants to use the bits explicitly on box (which may be the same or different from the toolset compiler, but are likely ngen'd), or when the user wants/needs to use an explicit copy of csc that may be located elsewhere (roslyn in bootstrap build, potentially other repos in source build, etc).

tmat:
> The option is still UsingToolMicrosoftNetCompilers, and it only opts out of using the desktop toolset compiler. This is desirable, both when a user wants to use the bits explicitly on box (which may be the same or different from the toolset compiler, but are likely ngen'd), or when the user wants/needs to use an explicit copy of csc that may be located elsewhere (roslyn in bootstrap build, potentially other repos in source build, etc).

I don't see the need for this property at this moment. Is it blocking dotnet SDK repo from migrating to RepoToolset? RepoToolset doesn't support source build yet. Once it does we may use a property like UsingToolMicrosoftNetCompilers, but maybe we will do something else. I don't want to add random properties like this without having source build working end-to-end. 

What about the UsingNetCoreCompilers compilers which would switch to the compilers included in dotnet SDK? I thought that's what's required by dotnet SDK repo.

tannergooding:
SDK repo supports building using both the CLI (default) or the full version of MSBuild (user opt-in on the command line).

Currently, the toolset compiler causes issues with XLIFF similar to the following:
```C#
D:\sdk\artifacts\Debug\obj\Microsoft.NET.Build.Tasks\netcoreapp1.0\cs\Microsoft.NET.Build.Tasks.resources.cs(10,30): error CS0433: The type 'AssemblyCompanyAttribute' exists in both 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' and 'System.Runtime, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' [D:\sdk\src\Tasks\Microsoft.NET.Build.Tasks\Microsoft.NET.Build.Tasks.csproj]
```

The easiest way to fix that is to opt-out of the toolset compiler.

It is an option we need now, and an option that is reasonable to have (and will likely be required) in the future. I see no reason to not have it.

tmat:
The toolset is not a kitchen sink of everything that might be potentially used in future. It's like public APIs or language features. Each feature starts with -1000 points. "I see no reason not to have it" doesn't increase the number of points a feature receives.

If there is an issue, we need to investigate it. Re this particular one: I saw this in the symreader-converter repo when using 2.0.* dotnet SDK. The root cause was a bug in the SDK that's fixed in 15.5.0-preview (converter uses 15.5.0-preview-007044).

jaredpar:
> The toolset is not a kitchen sink of everything that might be potentially used in future. It's like public APIs or language features.

I agree. I've seen a lack of discipline about features be the downfall over previous attempts at toolsets (many, many times). 

Maintenance is the biggest cost of any feature that is added here. It is orders o magnitude larger than the cost of initial implementation.

nguerrera:
We can't move repos in source-build forward to a build with that fix yet. Even if we did, we can't build xplat with a reference to Microsoft.Net.Compilers. Minimal option is: "do not override msbuild compiler on core".

tmat:
@nguerrera I thought we agreed on adding UsingNetCoreCompilers above that uses the dotnet SDK compiler? Does that not address the problem?

nguerrera:
@tmat Yes, it addresses the problem. I just got confused that maybe this was trending towards "we don't need anything". My bad.

tannergooding:
@tmat, @nguerrera. How is `UsingNetCoreCompilers` any different from what we have right now.

Since there isn't a netcore toolset compiler today, the props would be exactly the same, but with a different name.

tannergooding:
Ok, should all be resolved based on our offline discussions (@tmat, @nguerrera).

ProjectLanguage was removed, as it was likely a CLI/SDK specific thing copied from another repo

The option to disable using the toolset compiler will exist. Users will be expected to ensure this is deterministic whether in VS or not, but the long term goal should be that we have a single toolset compiler that contains both netcore and desktop implementations and references the correct one, depending on whether it is compiling under desktop msbuild or the .NET CLI.

The MSBuildAllProject definitions will be kept since they are best practice.

tannergooding:
@tmat: https://github.com/dotnet/roslyn-tools/issues/63

src/RepoToolset/ProjectDefaults.props
```diff
@@ -61,18 +68,18 @@
...
           Condition="'$(Language)' == '' and '$(MSBuildProjectExtension)' != '.shproj'"/>
-  
+
   <Choose>
     <!-- C# specific settings -->
     <When Condition="'$(Language)' == 'C#'">
       <PropertyGroup>
         <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
-        <NoWarn>$(NoWarn);1591</NoWarn>
+        <NoWarn Condition="'$(NoWarn)' == ''">1701;1702;1705;1591</NoWarn>
```
tmat: >Condition="'$(NoWarn)' == ''" [](start = 16, length = 29)

I don't think the condition is needed. These are defaults that the repo is expected to overwrite after they were defined.  The other properties in the file are also not conditional.

src/RepoToolset/RepoLayout.props
tmat: OK, let's keep this for now and we will update this when integrating with source build.

---
In reply to: [145773477](https://github.com/dotnet/roslyn-tools/pull/60#discussion_r145773477) [](ancestors = 145773477)

tmat:
👍

src/RepoToolset/ProjectDefaults.props
tannergooding: Fixed.

PR #61, opened 2017-10-19
adding PostCompileBinaryModification so the ApplyOptimizations task runs

jmarolf:
unless we write out a sentinel file the `ApplyOptimizations` task will never run.

image


tmat:
👍

PR #65, opened 2017-10-24
add auto-generated IVT file to F#'s CompileBefore collection

brettfo:
This ensures that the IVT files are included before the remaining code files which ensures that a program's entry point is never overwritten.


tmat:
👍

PR #71, opened 2017-11-01
Moving repo toolset to 1.0.0-alpha49

tannergooding:
FYI. @tmat, forgot to do this as part of the change yesterday.


tannergooding:
Also, where is the build definition that needs to be run to produce the new package version.

PR #77, opened 2017-11-05
Various test fixes for repo toolset.

tannergooding:
FYI. @tmat

### DIFF
--- a/src/RepoToolset/XUnit.targets
+++ b/src/RepoToolset/XUnit.targets
@@ -43,14 +43,14 @@
 
     <PropertyGroup Condition="'$(_RunOnCore)' == 'true'">
       <!-- TODO: xUnit doesn't support -html on CoreCLR currently (see https://github.com/xunit/xunit/issues/977) -->
-      <_TestRunnerCommand>"$(DotNetTool)" exec --depsfile "$(_CoreDepsPath)" --runtimeconfig "$(_CoreRuntimeConfigPath)" "$(NuGetPackageRoot)xunit.runner.console\$(XUnitVersion)\tools\netcoreapp1.0\xunit.console.dll" "$(TargetPath)" -noautoreporters -xml "$(_TestResultsXmlPath)"</_TestRunnerCommand>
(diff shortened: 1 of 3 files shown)


tannergooding:
This will let users manually pass in arguments they require, such as `-parallel none`.

tannergooding:
I also think this should be a general allowance, but not the long term fix for the parallel tests issue. This is because the optimal way to run tests in parallel is very dependent on the number of cores on the host machine.

For example, with the SDK repo, passing `-parallel none` and relying on test parallelization at the assembly level starts out using 87% of my processor at work, but only 21% of my processor at home. 

Additionally (in the SDK), since we have a couple assemblies that have significantly more tests than other assemblies, we finish out running tests using only 25% of the process at work, and only 6% of my processor at home.

Ideally, we should just be passing all assemblies into xUnit at once and let it parallelize them as it sees fit. This will parallelize both across and within assemblies, based on the number of cores in your machine, additionally taking into account the test attributes set by each test/assembly. This provides consistent utilization of your entire CPU for the duration, without overloading/underloading it with too many/few threads.

tannergooding:
Added a commit to consider `.Tests` as part of the 'UnitTestProject' set.

src/RepoToolset/XUnit.targets
```diff
@@ -43,14 +43,14 @@
 
     <PropertyGroup Condition="'$(_RunOnCore)' == 'true'">
       <!-- TODO: xUnit doesn't support -html on CoreCLR currently (see https://github.com/xunit/xunit/issues/977) -->
-      <_TestRunnerCommand>"$(DotNetTool)" exec --depsfile "$(_CoreDepsPath)" --runtimeconfig "$(_CoreRuntimeConfigPath)" "$(NuGetPackageRoot)xunit.runner.console\$(XUnitVersion)\tools\netcoreapp1.0\xunit.console.dll" "$(TargetPath)" -noautoreporters -xml "$(_TestResultsXmlPath)"</_TestRunnerCommand>
+      <_TestRunnerCommand>"$(DotNetTool)" exec --depsfile "$(_CoreDepsPath)" --runtimeconfig "$(_CoreRuntimeConfigPath)" "$(NuGetPackageRoot)xunit.runner.console\$(XUnitVersion)\tools\netcoreapp1.0\xunit.console.dll" "$(TargetPath)" -noautoreporters -xml "$(_TestResultsXmlPath) $(_TestRunnerAdditionalParameters)"</_TestRunnerCommand>
```
tmat: >_TestRunnerAdditionalParameters [](start = 281, length = 31)

I'd call this ```XUnitRunnerAdditionalArguments``` 
1) ```_``` prefixed properties are "private", this would be something repos should be setting, so it should not be private. 
2) The arguments are specific to XUnit. If we supported NUnit as well they would likely be different.

tannergooding: fixed.

PR #79, opened 2017-11-07
Updating _TestRunnerCommand so we don't get an error on Linux.

tannergooding:
FYI. @tmat 

Error was: dotnet exec needs a managed .dll or .exe extension. The application specified was '/home/tagoo/.nuget/packages/xunit.runner.console\2.3.0\tools\netcoreapp1.0\xunit.console.dll'


tmat:
👍

PR #83, opened 2017-11-14
Remove LocateVS

tmat:
Remove dependency on LocateVS from SignTool. The caller should pass msbuild path explicitly.


tmat:
@jaredpar @tannergooding

PR #102, opened 2017-11-26
Implement NuGetRepack tool

tmat:


jaredpar:
Is this the tool that allows for a NuPkg to be repackaged as the different flavors: release, perbuild release, etc ...?

src/NuGetRepack/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+        private static int Main(string[] args)
+        {
+            var operation = Operation.None;
+            var packages = new List<string>();
+            string outDirectory = null;
+
+            try
+            {
+                int i = 0;
+                while (i < args.Length)
```
jaredpar: Consider using Mono.options for the command line arguments. I've been using it for a lot of tools recently. Really simplifies the parsing and printing of help messages.

src/NuGetRepack/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+                }
+            }
+            catch (AggregateException e)
+            {
+                foreach (var inner in e.InnerExceptions)
+                {
+                    Console.Error.WriteLine(inner.Message);
+                }
+
+                return 3;
```
jaredpar: Consider giving names to magic numbers.

src/NuGetRepack/Program.cs
tmat: I don't think there is any benefit of adding extra dependency on Mono.Options. This is very trivial code. BTW, Mono.Options is not signed which is causing problems.

tmat:
Yes, that's exactly what it is.

src/NuGetRepack/Program.cs
tmat: ok.

PR #103, opened 2017-11-28
Updating RepoToolset and SignTool to support using dotnet build.

tannergooding:
FYI. @jaredpar, @tmat

### DIFF
--- a/src/SignTool/SignTool/SignTool.SignToolBase.cs
+++ b/src/SignTool/SignTool/SignTool.SignToolBase.cs
@@ -46,6 +46,11 @@ namespace SignTool
                 string fileName = MSBuildPath;
                 var commandLine = $@"/v:m ""{buildFilePath}""";
 
+                if (Path.GetFileNameWithoutExtension(fileName).Equals("dotnet", StringComparison.OrdinalIgnoreCase))
+                {
+                    commandLine = $"build {commandLine}";
+                }
+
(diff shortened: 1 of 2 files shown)


jaredpar:
What case do we need this for? 

Today my understanding all of the build machines have the desktop MSBuild installed. Is there a case where it doesn't?

src/SignTool/SignTool/SignTool.SignToolBase.cs
```diff
@@ -46,6 +46,11 @@ public void Sign(IEnumerable<FileSignInfo> filesToSign, TextWriter textWriter)
                 string fileName = MSBuildPath;
                 var commandLine = $@"/v:m ""{buildFilePath}""";
 
+                if (Path.GetFileNameWithoutExtension(fileName).Equals("dotnet", StringComparison.OrdinalIgnoreCase))
+                {
+                    commandLine = $"build {commandLine}";
+                }
```
jaredpar: Just thought through this and not sure this is going to work. The signing tool uses MSBuild tasks IIRC and it's unlikely they've been ported to target the non-desktop MSBuild.

tannergooding: Yeah, I talked with @tmat and he said the same thing. I've messaged you offline to try and determine if it is an issue with the pool or build definition.

PR #120, opened 2017-12-14
Updating SignTool to not require a strong name for assemblies.

tannergooding:


tannergooding:
@jaredpar for review.

Also, I was thinking that when you get the new signing validation up and working, we should have it run on `test`, for any assembly where `strongName` is `null`, since they should be getting signed with a full key there as well. Thoughts?

tannergooding:
FYI. @tmat on the version bump in repo-toolset.

The version will be correct, provided this is merged today and is included in the next signed build.

tmat:
If we want a change to be available we first need to make that change, kick off microbuild and then update the toolset to the nuget version that comes out of the build.

tmat:
👍

tannergooding:
ping @jaredpar, this is currently blocking.

PR #121, opened 2017-12-14
Trigger VAL builds in the same way as MicroBuild insertion task...

KevinH-MS:
Using BuildPolicy means the PR will be updated with status and links
from the DDRITs and RPS.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


KevinH-MS:
FYI @jasonmalinowski @tmeschter @cston 
/CC @dotnet/roslyn-infrastructure 

What's the best way to test out changes to the insertion tool?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -106,83 +108,80 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
             return newestBuild;
         }
 
-        private static async Task<Build> QueueValidationBuildAsync(string branchName)
+        // Similar to: https://devdiv.visualstudio.com/DevDiv/_git/PostBuildSteps#path=%2Fsrc%2FSubmitPullRequest%2FProgram.cs&version=GBmaster&_a=contents
+        private static async Task QueueBuildPolicy(GitPullRequest pullRequest, string buildPolicy)
         {
-            var buildClient = ProjectCollection.GetClient<BuildHttpClient>();
+            WaitForMergeComplete(pullRequest);
```
jaredpar: Seems weird to have a void blocking method inside an async one.

jmarolf: Yeah, but there is no async API to get this information

KevinH-MS: Not "seems"..."it is"... 😸

PR #122, opened 2017-12-20
Allow artifacts directory to be overridden

dsplaisted:
- Allow `ArtifactsDir` to be overridden.  This will let MSBuild build itself, and then use the built version of MSBuild to build itself again to
a different (stage 2) artifacts directory
- Add empty Directory.Build files in repo toolset targets folder.  This prevents Directory.Build.props and Directory.Build.targets being picked up when building Build.proj.
This allows you to set the `RepoToolsetDir` in another repo to a local repo for roslyn-tools to test out local changes.
- Show `.*proj` files (for example Build.proj) in solution explorer

### DIFF
--- a/sdks/RepoToolset/tools/Directory.Build.props
+++ b/sdks/RepoToolset/tools/Directory.Build.props
@@ -1,4 +1,5 @@
 <!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
 <Project>
-  
+  <!-- This is an empty Directory.Build.props file to prevent the Directory.Build.props file from the repo from being used
+       when building Build.proj if the repo toolset targets are imported directly from the repo -->
 </Project>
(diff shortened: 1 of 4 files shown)


tmat:
👍

sdks/RepoToolset/tools/Directory.Build.props
```diff
@@ -0,0 +1,4 @@
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project>
+
```
tmat: Add a comment why this file is needed

sdks/RepoToolset/tools/Directory.Build.targets
```diff
@@ -0,0 +1,4 @@
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project>
+
```
tmat: Add a comment why this file is needed

sdks/RepoToolset/tools/Directory.Build.props
jaredpar: Is the intent to stop the parent dir.build.props file from loading or to completely isolate this build? If it's the latter more is needed to cut off user customizations of build.

dsplaisted: The main intent is to stop the parent Directory.build files from being used, so that in another project using repo toolset you can do something like this:

```xml
<RepoToolsetDir>C:\git\roslyn-tools\sdks\RepoToolset\tools\</RepoToolsetDir>
```

And then you can test out changes to repo toolset locally.

PR #124, opened 2018-01-02
Fixing the official build version to not longer go over 65535

tannergooding:
FYI. @jaredpar, @jasonmalinowski, @tmat 

Port of https://github.com/dotnet/roslyn/pull/23999

### DIFF
--- a/sdks/RepoToolset/tools/Version.props
+++ b/sdks/RepoToolset/tools/Version.props
@@ -36,8 +36,18 @@
   <Choose>
     <When Condition="'$(OfficialBuild)' == 'true'">
       <PropertyGroup>
-        <_BuildNumberFiveDigitDateStamp>$([MSBuild]::Subtract($(BUILD_BUILDNUMBER.Split('.')[0].Substring(3).Trim()), 8800))</_BuildNumberFiveDigitDateStamp>
-        <_BuildNumberBuildOfTheDayPadded>$(BUILD_BUILDNUMBER.Split('.')[1].PadLeft(2,'0'))</_BuildNumberBuildOfTheDayPadded>
+        <!-- Split the build parts out from the BuildNumber which is given to us by MicroBuild in the format of yyyymmdd.nn
+         where BuildNumberFiveDigitDateStamp is mmmdd (such as 60615) and BuildNumberBuildOfTheDay is nn (which represents the nth build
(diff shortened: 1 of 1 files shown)


tannergooding:
Looks like this actually needs to be tweaked slightly. It apparently breaks cross-plat msbuild according to @jasonmalinowski

sdks/RepoToolset/tools/Version.props
```diff
@@ -36,8 +36,18 @@
   <Choose>
     <When Condition="'$(OfficialBuild)' == 'true'">
       <PropertyGroup>
-        <_BuildNumberFiveDigitDateStamp>$([MSBuild]::Subtract($(BUILD_BUILDNUMBER.Split('.')[0].Substring(3).Trim()), 8800))</_BuildNumberFiveDigitDateStamp>
-        <_BuildNumberBuildOfTheDayPadded>$(BUILD_BUILDNUMBER.Split('.')[1].PadLeft(2,'0'))</_BuildNumberBuildOfTheDayPadded>
+        <!-- Split the build parts out from the BuildNumber which is given to us by MicroBuild in the format of yyyymmdd.nn
```
tmat: Perhaps it's time to define a task and write this logic in C# instead of this mess of msbuild calls...

tmat:
👍 To unblock builds.

sdks/RepoToolset/tools/Version.props
tannergooding: Maybe (and to update the logic so that it doesn't require updating every year or two, since this new logic will still fail at some point in the future). But that is probably a future change since everything is on the floor right now.

jasonmalinowski: @tannergooding How will the new logic fail (other than actually running out of numbers which obviously requires you to restart)

tannergooding: Its still a general issue and we have to ensure that resetting coincides with a increment of one of the other version numbers.

jasonmalinowski:
@tmat and @tannergooding I've updated the Roslyn one with the correct version; you might want to take it assuming it also applies to here too.

tannergooding:
The logic is already contained to `OfficialBuild` here, so we don't have the same issue that Roslyn had for DevBuilds.

PR #132, opened 2018-01-09
Make GetBuildVersion target optional to avoid conflict with Nerdbank.GitVersioning

dsplaisted:
The [Nerdbank.GitVersioning](https://github.com/AArnott/Nerdbank.GitVersioning) package defines a `GetBuildVersion` target: https://github.com/AArnott/Nerdbank.GitVersioning/blob/d398833fcdcb857fe44c3e8287f25b4058921181/src/Nerdbank.GitVersioning.NuGet/build/Nerdbank.GitVersioning.targets#L62

This conflicts with the `GetBuildVersion` target which repo toolset defines if `UsingToolVSSDK` is set to true.  The result is that if you try to enable the VS SDK in a project, then Nerdbank.GitVersioning will stop working.

This PR adds a workaround for this: you can set the `DefineGetBuildVersionTargetForVsix` to false to prevent repo toolset's version of this target from being defined.

I ran into this while [porting MSBuild to use repo toolset](https://github.com/Microsoft/msbuild/issues/2706).

FYI @AArnott

### DIFF
--- a/sdks/RepoToolset/tools/GetBuildVersion.targets
+++ b/sdks/RepoToolset/tools/GetBuildVersion.targets
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project>
+  
+  <!-- Returns the current build version. Used in .vsixmanifests to substitute our build version into them.
+      This target is defined in its own file so that it can be conditionally defined.  This is so it is
+      possible to avoid conflicts with https://github.com/AArnott/Nerdbank.GitVersioning, which also
+      defines a GetBuildVersion target.
+  -->
+  <Target Name="GetBuildVersion" Outputs="$(VsixVersion)" />
+
+</Project>
--- a/sdks/RepoToolset/tools/VisualStudio.targets
+++ b/sdks/RepoToolset/tools/VisualStudio.targets
@@ -243,7 +243,15 @@
     <GetVsixSourceItemsDependsOn>$(GetVsixSourceItemsDependsOn);_GetVsixTemplateItems</GetVsixSourceItemsDependsOn>
   </PropertyGroup>
 
-  <!-- Returns the current build version. Used in .vsixmanifests to substitute our build version into them -->
-  <Target Name="GetBuildVersion" Outputs="$(VsixVersion)" />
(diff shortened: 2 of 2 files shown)


tmat:
Why do we need Nerdbank.GitVersioning? The toolset provides git commit sha: https://github.com/dotnet/roslyn-tools/blob/master/sdks/RepoToolset/tools/GitHash.props

tmat:
Also, in future we will replace this with https://github.com/tmat/repository-info/tree/master/docs

dsplaisted:
@tmat MSBuild already uses Nerdbank.GitVersioning and doesn't want to have to change their version format when adopting repo toolset.

tmat:
@dsplaisted What's their version format? We are going to need to change this eventually.

tmat:
The point of RepoToolset is to have a uniform infrastructure. This is part of the infrastructure.

tmat:
That said, we should probably rename ```GetBuildVersion``` to ```GetVsixVersion```, but that will require updates across multiple repos. Could you please file an issue and link it to your change?
We can take your change as a temporary workaround. Also I'd file a bug on msbuild to stop using Nerdbank.GitVersioning once the functionality is available in msbuild.

AArnott:
> we should probably rename GetBuildVersion to GetVsixVersion

Agreed.

> file a bug on msbuild to stop using Nerdbank.GitVersioning

I'll be interested in that discussion. NB.GV does a lot more than the sha lookup that this toolset does. I wonder what you have against using NB.GV when it works well for so many repos already.

tmat:
>  I wonder what you have against using NB.GV when it works well for so many repos already.

Nothing at all when used by other repos. However, we are unifying *our* repos under a single infrastructure toolset. It is highly beneficial for all these repos to agree on the same versioning and infra tooling.

AArnott:
> It is highly beneficial for all these repos to agree on the same versioning and infra tooling

Fair enough. Have you considered standardizing on NB.GV instead of removing it? I'd be happy to chat offline about the value NB.GV offers to see if it sounds good to you. I know msbuild thought it was worthwhile when they added it.

AArnott:
>> I wonder what you have against

My word choice implied I was accusing @tmat of personal bias against a technology. Sorry about that. I didn't mean to.

tmat:
@AArnott No worries. No harm done :). Standardizing on NB.GV is certainly an option if it adds enough value that's worth adding extra dependency and making the set of tools needed to build all our repos bigger. I'm somewhat skeptical, but could be persuaded. I'm sure whoever used it in msbuild thought it adds some value. However, it's usually easier to keep adding than to remove.

PR #137, opened 2018-01-12
Use Roslyn instead of al.exe to generate satellite assemblies

dsplaisted:
Use Satellite assembly generation task from Microsoft.NET.Sdk even when building with full Framework MSBuild.  This will support public signing, is deterministic, and always generates them as AnyCPU.


tmat:
👍

PR #139, opened 2018-01-17
Updating PipeBuild.Publish to pass along manifest information.

tannergooding:
Just a first pass doing nearly the simplest thing.

### DIFF
--- a/sdks/RepoToolset/tools/PipeBuild.Publish.proj
+++ b/sdks/RepoToolset/tools/PipeBuild.Publish.proj
@@ -2,8 +2,8 @@
 <Project DefaultTargets="Publish">
   <!--
     Required variables:
-      VersionsPropsPath          Versions.props path. 
-      FixedVersionsPropsPath     FixedVersions.props path. 
+      VersionsPropsPath          Versions.props path.
+      FixedVersionsPropsPath     FixedVersions.props path.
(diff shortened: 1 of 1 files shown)


tmat:
Why don't we just use the environment variables? The publishing is only relevant in official builds where the variables are always available.

tannergooding:
> Why don't we just use the environment variables? T

We don't use them directly in order to more easily support other environments or future directions where we don't use a build system that has those variablese.

tmat:
I'd add those when that will happen. Adding them now makes things unnecessarily complex.

tannergooding:
Updated.

tmat:
Thanks!

dagood:
I think this is missing a Feed package upgrade:

https://devdiv.visualstudio.com/DevDiv/_build?buildId=1301675

> error MSB4064: The "ManifestName" parameter is not supported by the "PushToBlobFeed" task.

tannergooding:
@dagood, It is currently targeting `2.1.0-prerelease-02221-02` (https://github.com/dotnet/roslyn-tools/blob/1658d012a482a0caaff7a58a0dc941bb417edcd3/sdks/RepoToolset/tools/DefaultVersions.props#L41).

What is the correct version to target?

PR #141, opened 2018-01-19
Updating the .NET Build Feed task to 2.1.0-prerelease-02419-01

tannergooding:
FYI. @tmat, @dagood


dagood:
I looked at the BuildTools commits and it looks like a breaking change came in: https://github.com/dotnet/buildtools/pull/1868. You'll have to either:

 * Use an earlier version (say, `2.1.0-prerelease-02417-06`)
 * (Recommended to get over the hump) Add an `Import` on the props file as described here: https://github.com/dotnet/buildtools/pull/1868#issuecomment-359036782

tannergooding:
Added an import of the props file.

tmat:
👍 Thanks

PR #143, opened 2018-01-22
Updating RepoToolset to force public signing on non-windows

tannergooding:
FYI. @jaredpar, @tmat

### DIFF
--- a/sdks/RepoToolset/tools/StrongName.targets
+++ b/sdks/RepoToolset/tools/StrongName.targets
@@ -50,6 +50,11 @@
     </When>
   </Choose>
 
+  <!-- Full signing is not currently supported on non-Windows -->
+  <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
+    <PublicSign>true</PublicSign>
+  </PropertyGroup>
+
   <!-- Build Flag Verification -->
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


sdks/RepoToolset/tools/StrongName.targets
```diff
@@ -50,6 +50,11 @@
     </When>
   </Choose>
 
+  <!-- Full signing is not currently supported on non-Windows -->
```
tmat: I believe the compilers already support it. Shall we move to the newer compilers version instead?

tannergooding: Some reps aren't using the tool set compilers and are using a particular cli build instead.

tannergooding: Not in all cases. For something like release/2.0.0, we are explicitly building with a 2.0 build of the cli.

tmat: Then I'd say the repo should set this. We want to use the latest compilers and features by default.

jaredpar: I agree.

IMHO: The lag of toolset adoption means that either RepoToolset or individual repos need to tweak certain settings. That will be true for this change and others in the future. I would rather make it more painful for repos who are not adopting completely compatible toolsets than add too much complexity to the central tools.

PR #144, opened 2018-01-24
Remove wait for merge status...

KevinH-MS:


KevinH-MS:
FYI @jasonmalinowski @dotnet/roslyn-infrastructure 

This code was copied from the MicroBuild action, but the insertion tool kept timing out waiting for this status.  I tested a couple times, and in practice, there does not appear to be any harm in just kicking off the DDRITS/RPS as soon as the PR is created.

KevinH-MS:
@jasonmalinowski also FYI, I went ahead and proactively rolled out this change on roslyn-svcs.

KevinH-MS:
Today's insertions (with this change) all look good.

PR #145, opened 2018-01-25
Added PrivateAssets="all" to development dependencies

jp2masa:
## Changes
- Added PrivateAssets="all" to development dependencies.
- Updated RepoToolset.md.

### DIFF
--- a/docs/RepoToolset.md
+++ b/docs/RepoToolset.md
@@ -99,7 +99,7 @@ Versions.props file is present in the repo and defines versions of all dependenc
     <UsingToolIbcOptimization>true</UsingToolIbcOptimization>
   
     <!-- RepoToolset package version -->
-    <RoslynToolsMicrosoftRepoToolsetVersion>1.0.0-alpha27</RoslynToolsMicrosoftRepoToolsetVersion>
+    <RoslynToolsRepoToolsetVersion>1.0.0-alpha27</RoslynToolsRepoToolsetVersion>
     
     <!-- Tool versions when using dotnet cli build driver -->
     <DotNetCliVersion>1.0.0-rc4-004777</DotNetCliVersion>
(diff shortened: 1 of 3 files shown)


docs/RepoToolset.md
```diff
@@ -99,7 +99,7 @@ Versions.props file is present in the repo and defines versions of all dependenc
     <UsingToolIbcOptimization>true</UsingToolIbcOptimization>
   
     <!-- RepoToolset package version -->
-    <RoslynToolsMicrosoftRepoToolsetVersion>1.0.0-alpha27</RoslynToolsMicrosoftRepoToolsetVersion>
+    <RoslynToolsRepoToolsetVersion>1.0.0-alpha27</RoslynToolsRepoToolsetVersion>
```
sharwell: :memo: The version itself seems to be old. Not sure if it needs to be updated now or later.

sdks/RepoToolset/tools/Compiler.props
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Net.Compilers" Version="$(MicrosoftNetCompilersVersion)" />
+    <PackageReference Include="Microsoft.Net.Compilers" Version="$(MicrosoftNetCompilersVersion)" PrivateAssets="all" />
```
jaredpar: Can we elaborate on what was breaking here? I agree this is the right choice but also want to understand what impact this had (can't think of a prob it creates).

jp2masa: `Microsoft.Net.Compilers` is a development dependency, but `PackageReference` doesn't have support for `developmentDependency` metadata yet. Currently, packages created from project files (like `Microsoft.VisualStudio.ProjectSystem.Managed`) pull `Microsoft.Net.Compilers`, but it shouldn't be included as a package dependency, as it's only used to build the projects.

tmat:
👍 Thanks!

PR #149, opened 2018-01-31
Fix more signing problems

agocke:


sdks/RepoToolset/tools/StrongName.targets
```diff
@@ -51,7 +51,7 @@
   </Choose>
 
   <!-- Full signing is not currently supported on non-Windows -->
-  <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
+  <PropertyGroup Condition="'$(OS)' != 'Windows_NT' and '$(SignAssembly)' != 'false'">
     <PublicSign>true</PublicSign>
```
jaredpar: Why is this group even needed? PublicSign is already set to true based on the SignAssembly property above.

agocke: It's probably not... I'll fix this in a subsequent PR.

PR #152, opened 2018-02-07
Add the Github merge bot source code

agocke:
Also implements a small change to check if there's an existing PR and
skip PR creation if so.


agocke:
@jasonmalinowski This change should get us to stop making duplicate merge PRs. I have a follow-up change that enables auto-merging. Does `dotnet-bot` have admin yet?

agocke:
@sharwell I rebased it into two commits. Is that good enough?

src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -38,32 +39,49 @@ public class GithubMergeTool
...
             string srcBranch,
             string destBranch)
         {
+            string prTitle = $"Merge {srcBranch} to {destBranch}";
+
+            // Check to see if there's already a PR
+            // TODO: handle pagination
+            // This handles the 100 most recent PRs, which should work for now
+            HttpResponseMessage prsResponse = await _client.GetAsync(
+                $"repos/{repoOwner}/{repoName}/pulls?state=open&base={destBranch}&per_page=100");
```
sharwell: ❗️ This doesn't check if the existing PR is up-to-date.

:memo: Keep in mind that when you merge the most recent PR for a target branch, all other open PRs for that same target branch close automatically. The previous approach created more pull requests, but it allowed for seamless review using either an incremental approach or by simply taking the most recent one.

I would recommend simply not including this feature.

jasonmalinowski: @sharwell The creation of multiple PRs created lots of confusion and duplicate work. Furthermore, with this and the auto-merge support we're planning on running this once an _hour_ not once a day, which would make duplicates a non-starter.

jasonmalinowski: And some background: right now doing these merges once a day is proving to be a significant bottleneck on code flow. The extra day of latency causes problems with shiproom, ensuring we don't have insertion conflicts between 15.6/15.7, etc.

src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -0,0 +1,150 @@
...
+            string srcBranch,
+            string destBranch)
+        {
+            string prTitle = $"Merge {srcBranch} to {destBranch}";
+
+            // Check to see if there's already a PR
+            // TODO: handle pagination
+            // This handles the 100 most recent PRs, which should work for now
+            HttpResponseMessage prsResponse = await _client.GetAsync(
+                $"repos/{repoOwner}/{repoName}/pulls?state=open&base={destBranch}&per_page=100");
```
jasonmalinowski: Can this also filter to ones being made by dotnet-bot? That might help address the pagination problem.

src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -0,0 +1,150 @@
...
+        public async Task<(bool prCreated, HttpResponseMessage error)> CreateMergePr(
+            string repoOwner,
+            string repoName,
+            string srcBranch,
+            string destBranch)
+        {
+            string prTitle = $"Merge {srcBranch} to {destBranch}";
+
+            // Check to see if there's already a PR
+            // TODO: handle pagination
```
jasonmalinowski: We're too good to use Octokit? :-)

src/GithubMergeTool/GithubMergeTool.cs
sharwell: @jasonmalinowski Thanks, that's a sufficient response for me considering I'm not Tiger.

src/GithubMergeTool/GithubMergeTool.cs
sharwell: I was just coming back to say this.

agocke: Not a parameter in the `pulls` API.

src/GithubMergeTool/GithubMergeTool.cs
agocke: Dependencies in Azure functions are tricky. I'd rather not add more.

src/GithubMergeTool/GithubMergeTool.cs
agocke: Wait, I think I misread something. Maybe.

sharwell: This works for me (only shows PR from me): https://api.github.com/repos/Microsoft/perfview/pulls?author=sharwell

src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -52,12 +54,13 @@ public class GithubMergeTool
...
         {
             string prTitle = $"Merge {srcBranch} to {destBranch}";
+            string prBranchName = $"merges/{srcBranch}-to-{destBranch}";
 
             // Check to see if there's already a PR
             // TODO: handle pagination
             // This handles the 100 most recent PRs, which should work for now
             HttpResponseMessage prsResponse = await _client.GetAsync(
-                $"repos/{repoOwner}/{repoName}/pulls?state=open&base={destBranch}&per_page=100");
+                $"repos/{repoOwner}/{repoName}/pulls?state=open&base={destBranch}&per_page=100&head={_username}:{prBranchName}");
```
sharwell: ❔  With the additional filters, can you limit this to a single item.

src/GithubMergeTool/GithubMergeTool.cs
agocke: This doesn't appear to be documented: https://developer.github.com/v3/pulls/#list-pull-requests

What *is* documented is searching by both the author and branch name. Since we will now only have one merge PR for each two branches, I can remove the date from the branch name and make it predictable. I've done so in the latest commit.

src/GithubMergeTool/GithubMergeTool.cs
agocke: I could, but don't see the benefit.

agocke: I will, however, remove the comment. That is confusing.

src/GithubMergeTool/GithubMergeTool.cs
jasonmalinowski: @agocke Just make sure that if somebody closes (but does not merge) the PR we are able to recover cleanly. I could imagine we screw up a merge so badly somebody just closes the PR and might not realize it's not working again.

src/GithubMergeTool/GithubMergeTool.cs
agocke: Yeah, I'm removing the explicit 100 count request and the comment. That seems like the simplest, most straightforward API call, and the logic on the bot side seems simple as well.

src/GithubMergeTool/GithubMergeTool.cs
agocke: Good point. If someone merges the PR but forgets to delete the branch, that may block a new merge from being created. Would you be OK with the bot force-resetting the branch?

src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -94,11 +94,31 @@ public class GithubMergeTool
     ""sha"": ""{srcSha}""
 }}";
 
+            Console.WriteLine("Creating branch");
             response = await _client.PostAsyncAsJson($"repos/{repoOwner}/{repoName}/git/refs", body);
 
             if (response.StatusCode != HttpStatusCode.Created)
             {
-                return (false, response);
+                // Branch already exists. Hard reset to the new SHA
```
sharwell: ❓ What if the branch exists, and has been manually updated to address a merge conflict?

sharwell: ➡️ The answer is in the commit message. For this case, there is a branch but no pull request.

agocke: If the PR is open then we should have bailed out earlier in this method.

Issue #151, opened 2018-02-06
Insertion tool should only try to insert standalone builds


jasonmalinowski:
Right now the insertion tool is trying to insert builds kicked off as part of the overall product construction build. It should only be trying to insert builds that actually have artifacts on our drop share.

AbhitejJohn:
@jasonmalinowski : please feel free to close this one. We've already merged the change.

PR #155, opened 2018-02-07
Fixed up the logic to only pick builds with a valid artifact.

AbhitejJohn:


AbhitejJohn:
Fixes #151.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -60,9 +60,11 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
             var buildClient = ProjectCollection.GetClient<BuildHttpClient>();
             var definitions = await buildClient.GetDefinitionsAsync(project: Options.TFSProjectName, name: Options.BuildQueueName);
             var builds = await GetBuildsFromTFSAsync(buildClient, definitions, cancellationToken, BuildResult.Succeeded);
+
+            // Get the latest build with valid artifacts.
             return (from build in builds
                     orderby build.FinishTime descending
-                    select build).FirstOrDefault();
+                    select build).FirstOrDefault((b)  => (buildClient.GetArtifactsAsync(b.Project.Id, b.Id, cancellationToken).Result.Any(a => !string.IsNullOrEmpty(a.Name) && a.Name.Contains(b.BuildNumber))));
```
jasonmalinowski: It might be safer to do the .Contains() and say something like "contains \\cpvsbuild". I could imagine artifacts ending up somewhere else with the name but we can't consume it that way...?

jasonmalinowski: Might be a bit less messy to make this a where clause in the LINQ query.

PR #158, opened 2018-02-12
Enabling repo-defined performance test execution.

ghost:
Intended implementation:
When the Boolean "PerformanceTest" is false, test projects with the naming convention: *.PerformanceTests run as normal tests. When "PerformanceTest" is true, call the repo built performance test harness.

@dotnet/dotnet-cli


sdks/RepoToolset/tools/Tests.props
```diff
@@ -6,6 +6,11 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(IsPerformanceTestProject)' == ''">
+    <IsPerformanceTestProject>false</IsPerformanceTestProject>
+    <IsPerformanceTestProject Condition="$(MSBuildProjectName.EndsWith('.PerfTests')) or $(MSBuildProjectName.EndsWith('.PerformanceTests'))">true</IsPerformanceTestProject>
```
tannergooding: Do we need to support both `PerfTests` and `PerformanceTests`?

johnbeisner: Why would we not do this?

tannergooding: It is easier to say there is one recognized pattern and to stick to it. `PerfTests` seems sufficient enough.

sdks/RepoToolset/tools/Performance.targets
```diff
@@ -0,0 +1,23 @@
...
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+  </PropertyGroup>
+
+  <Target Name="PerformanceTest" Condition="'$(IsPerformanceTestProject)' == 'true'">
+
+    <MakeDir Directories="$(ArtifactsTestResultsDir)"/>
+    <Message Text="Running performance tests: $(TargetPath)" Importance="high"/>
+
+    <Exec Command='$(DotNetTool) $(TargetPath) $(PerformanceTestParameters)'
```
tmat: >$(DotNetTool) $(TargetPath) [](start = 19, length = 27)

Missing quotes:
`Command="$(DotNetTool)" "$(TargetPath)" $(PerformanceTestParameters)'`

johnbeisner: Done...

sdks/RepoToolset/tools/Performance.targets
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project>
+
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+  </PropertyGroup>
+
+  <Target Name="PerformanceTest" Condition="'$(IsPerformanceTestProject)' == 'true'">
```
tmat: ><Target Name="PerformanceTest"  [](start = 2, length = 31)

The target will need to be defined in XUnit.targets as well.
See [https://github.com/dotnet/roslyn-tools/blob/master/sdks/RepoToolset/tools/XUnit.targets#L21](https://github.com/dotnet/roslyn-tools/blob/master/sdks/RepoToolset/tools/XUnit.targets#L21)

johnbeisner: Done...

sdks/RepoToolset/tools/Tests.props
johnbeisner: Okay...

sdks/RepoToolset/tools/Performance.targets
tmat: The target has to be defined for all projects, so that we can run `msbuild /t:PerformanceTest Foo.sln`. For non-perf-test project the target needs to be no-op.

---
In reply to: [167685658](https://github.com/dotnet/roslyn-tools/pull/158#discussion_r167685658) [](ancestors = 167685658)

sdks/RepoToolset/tools/Performance.targets
tmat: >PerformanceTestParameters [](start = 49, length = 25)

Where do these parameters come from?

tmat: Are they defined by the project?

---
In reply to: [167686054](https://github.com/dotnet/roslyn-tools/pull/158#discussion_r167686054) [](ancestors = 167686054)

tmat:
> When the Boolean "PerformanceTest" is false, test projects with the naming convention: *.PerfTests or .PerformanceTests run as normal tests

Does this mean they are regular XUnit tests?

johnbeisner:
@tmat 
They are intended to run as regular xunit functional tests and 'repo-specific-test-harness' performance tests.

tmat:
Also are these tests only runnable on .NET Core? Do we not have a runner for .NET Framework?

tmat (approved):
:shipit:

tannergooding (approved):

PR #159, opened 2018-02-14
Make RepoToolset an SDK package

tmat:
The required repository layout now looks like so:
- All projects in the repo shall start with `<Project Sdk="RoslynTools.RepoToolset">`

- `/global.json`

```json
{
  "sdk": {
    "version": "2.1.100-preview-007366"   // dotnet cli/sdk version to use, must support custom sdks
  },
  "msbuild-sdks": {
    "RoslynTools.RepoToolset": "1.0.0-xxx"   // repo toolset version to use
  }
}
```

- `/nuget.config`

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <!-- Only specify feed for RepoToolset SDK (see https://github.com/Microsoft/msbuild/issues/2982) -->
  <packageSources>
    <add key="roslyn-tools" value="https://dotnet.myget.org/F/roslyn-tools/api/v3/index.json" />
  </packageSources>
</configuration>
```

- `/build/Versions.props`

Versions of dependencies and the artifacts produced by the repo, source feeds for dependencies specified via `RestoreSources` property. No longer includes dotnet CLI or RepoToolset versions.

- `/build/SignToolData.json`

Signing info.

- `/build/build.ps1` Windows bootstrapper script: [example](https://github.com/tmat/symreader-converter/blob/b7499daa55a207ce253e33c4664f8532d515e712/build/build.ps1). 

- `Directory.Build.props/targets` no longer import RepoToolset props/targets.
- `Directory.Build.props/targets` are no longer required anywhere in the repo (still encouraged to share common props/targets across projects).

Use case: https://github.com/dotnet/symreader-converter/pull/117

A couple of workarounds needed for:
https://github.com/Microsoft/msbuild/issues/2992
https://github.com/Microsoft/msbuild/issues/2982


tmat:
@nguerrera @tannergooding FYI

tmat:
@markwilkie @natemcmaster FYI

tmat:
@jinujoseph FYI

sdks/RepoToolset/tools/Build.proj
```diff
@@ -185,15 +176,6 @@
              BuildInParallel="true"
              Condition="'@(_SolutionBuildTargets)' != ''" />
 
-    <!--
-      Workaround for https://github.com/NuGet/Home/issues/6165.
```
chcosta: I presume that this workaround will still be needed if the toolset is going to (potentially) be consumed in a repo using an older version of the NuGet client.  Probably not an issue for RepoToolset repos but it may be for the arcade tools?

tmat: Not sure what do you mean. This build project is only gonna be used in repos using RepoToolset. It implements the CI build script for these repos. Also I don't think we should support older NuGet clients in any of our repos. They should all be on the latest released build.

chcosta: The comment was meant merely to confirm my assertion that the problem has been fixed in newer versions of the NuGet client, but if someone attempted to use RepoToolset, or some other toolset with similar semantics, in a servicing branch, then the problem would still exist.  It was a general comment and not specifically applicable to this PR.  

Thanks for making the SDK change!

tmat: Oh, I see. Yes, it would exists in older branches.

PR #164, opened 2018-02-15
Trigger a release to symbol archive.

AbhitejJohn:
Added ability to trigger a release before creating an insertion PR so symbol archival requests are made.
Also updated everything to the latest packages.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
@@ -15,13 +16,19 @@ using Microsoft.TeamFoundation.Client;
 using Microsoft.TeamFoundation.Policy.WebApi;
 using Microsoft.TeamFoundation.SourceControl.WebApi;
 using Microsoft.VisualStudio.Services.Common;
-
+using Microsoft.VisualStudio.Services.ReleaseManagement.WebApi;
+using Microsoft.VisualStudio.Services.ReleaseManagement.WebApi.Clients;
+using Microsoft.VisualStudio.Services.ReleaseManagement.WebApi.Contracts;
+using Microsoft.VisualStudio.Services.WebApi;
 using Newtonsoft.Json;
 
 namespace Roslyn.Insertion
 {
     static partial class RoslynInsertionTool
     {
+        // Currently this is the only release we trigger. We can easily move this over to options when we need this configurable.
+        private const string ReleaseDefinitionName = "Roslyn";
+
         private static readonly Lazy<TfsTeamProjectCollection> LazyProjectCollection = new Lazy<Microsoft.TeamFoundation.Client.TfsTeamProjectCollection>(() =>
         {
             Log.Trace($"Creating TfsTeamProjectCollection object from {Options.VSTSUri}");
@@ -50,6 +57,8 @@ namespace Roslyn.Insertion
             return await gitClient.CreatePullRequestAsync(
                     CreatePullRequest("refs/heads/" + branchName, "refs/heads/" + Options.VisualStudioBranchName, message),
                     repository.Id,
+                    supportsIterations: null,
+                    userState: null,
                     cancellationToken);
         }
 
@@ -60,7 +69,7 @@ namespace Roslyn.Insertion
             var buildClient = ProjectCollection.GetClient<BuildHttpClient>();
             var definitions = await buildClient.GetDefinitionsAsync(project: Options.TFSProjectName, name: Options.BuildQueueName);
             var builds = await GetBuildsFromTFSAsync(buildClient, definitions, cancellationToken, BuildResult.Succeeded);
-
+            
             // Get the latest build with valid artifacts.
             return (from build in builds
                     where buildClient.GetArtifactsAsync(build.Project.Id, build.Id, cancellationToken).Result.Any(a => a.Resource != null && a.Resource.Data != null && a.Resource.Data.Contains(Options.BuildDropPath))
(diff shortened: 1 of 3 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -15,13 +16,19 @@
...
+using Microsoft.VisualStudio.Services.ReleaseManagement.WebApi.Contracts;
+using Microsoft.VisualStudio.Services.WebApi;
 using Newtonsoft.Json;
 
 namespace Roslyn.Insertion
 {
     static partial class RoslynInsertionTool
     {
+        // Currently this is the only release we trigger. We can easily move this over to options when we need this configurable.
+        private const string ReleaseDefinitionName = "Roslyn";
```
jasonmalinowski: So we will also need the equivalent for LUT and Project System. So we'll need this configurable right away.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -181,6 +171,136 @@ where buildClient.GetArtifactsAsync(build.Project.Id, build.Id, cancellationToke
...
+            var releaseMetadata = new ReleaseStartMetadata()
+            {
+                DefinitionId = releaseDefinition.Id,
+                Description = $"Automated release for {Options.BranchName} for build {build.BuildNumber}",
+                Reason = ReleaseReason.ContinuousIntegration
+            };
+
+            var artifactMetadata = new ArtifactMetadata
+            {
+                Alias = "Roslyn-Signed", // This is the alias in the Release definition.
```
jasonmalinowski: This should pull in the name of whatever build queue we kicked off, right?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -181,6 +171,136 @@ where buildClient.GetArtifactsAsync(build.Project.Id, build.Id, cancellationToke
...
+            }
+
+            Log.Info("GetReleaseEnvironmentStatus: Getting release environment status for environmentId: {0}, ReleaseId:{1}", environmentId, release.Id);
+
+            var environment = GetReleaseEnvironment(projectName, release.Id, environmentId, releaseClient, cancellationToken);
+
+            return environment.Status;
+        }
+
+        private static ReleaseEnvironment GetReleaseEnvironment(string projectName, int releaseId, int releaseEnvironmentId, ReleaseHttpClient releaseClient, CancellationToken cancellationToken)
```
jasonmalinowski: Not sure why this method exists...it's just forwarding?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -181,6 +171,136 @@ where buildClient.GetArtifactsAsync(build.Project.Id, build.Id, cancellationToke
...
+                if (releaseEnvironment.Status != EnvironmentStatus.Succeeded)
+                {
+                    Log.Error($"The release did not succeed. Take a look at {release.ReleaseDefinitionReference.Url} to find {release.Name} for more details.");
+                }
+            }
+
+            return releaseEnvironment.Status;
+        }
+
+        private static EnvironmentStatus GetReleaseEnvironmentStatus(string projectName, Release release, int environmentId, ReleaseHttpClient releaseClient, CancellationToken cancellationToken)
```
jasonmalinowski: Not sure why this method exists...it's just forwarding?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -202,6 +202,20 @@ public static partial class RoslynInsertionTool
                     }
                 }
 
+                // ********************* Trigger a release *****************************
+                Log.Info($"Triggering a release for the build {newestBuild.BuildNumber}");
+
+                var release = await CreateReleaseAsync(newestBuild, cancellationToken);
+
+                // The timeout for the below wait is primarily dependent on:
```
jasonmalinowski: I honestly wouldn't wait more than 5 minutes or so. It's more important for us to get the PR going so we can start val builds than anything else, I'd think.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
AbhitejJohn: The same release definition works for LUT and project system too. So we should be good. Will add comments appropriately. Any suggestions for a name change?

AbhitejJohn: How about `ManagedLanguages`?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
AbhitejJohn: This is the alias for the release definition specified in the release definition. It isn't tied to the build queue.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
AbhitejJohn: I can reduce this timeout to 5/10 minutes but it would fail in cases when the build queue is swamped with other requests already. Two ways of getting out that though:
1. Have a dedicated always available pool.
2. Figure out a way of kicking of the release definition way earlier when the build itself is complete. We could achieve this selectively via tags. Since the build triggered by vslsnap would mostly be triggered once a day(during the insertion changes) we wouldn't be swamping the symbol archival. That could also enable others to selectively auto-trigger release when required.

AbhitejJohn: Reduced the timeout. Will monitor this over the next few days to see if we run into any spikes like I saw yesterday night.

AbhitejJohn: And the symbol check seems to be auto-triggering today(which does make sense), so it looks like there isn't a way to delay waiting for the release to complete after creating a PR.

AbhitejJohn:
@jasonmalinowski for a second iteration.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
jasonmalinowski: @tmat Do you already have a standard one used for all repo-tools based repos?

tmat:
@jasonmalinowski Are you sure we want to do release on every insertion? Sounds like overkill.

The CLR is now publishing symbols directly to MSDL server using their new publishing tool. Some changes are needed in order for the tool to be usable for Roslyn and other repos. Once those are done we can simplify our symbol publishing significantly.

jasonmalinowski:
@tmat VS release is requiring that inserted bits are archived. Any new insertion that doesn't do this creates bugs, which we have about 100 of right now assigned to me.

Sure, I'd be OK with that new tool if/when it exists. Until then, we need this desperately.

tmat:
We can white-list our binaries. I already did so, but somehow it got lost in merges.

tmat:
If this works well then i'm fine with doing it for now.

PR #165, opened 2018-02-25
Update the workaround for xunit/xunit#1651

sharwell:
* Works with the current xunit release
* Works with fast up-to-date check in Visual Studio


tmat:
Thanks!

PR #169, opened 2018-02-26
Switch PowerShell to TLS1.2

tmat:


build/Versions.props
```diff
@@ -8,7 +8,7 @@
     <!-- Toolset -->
     <RoslynToolsRepoToolsetVersion>1.0.0-beta-62614-04</RoslynToolsRepoToolsetVersion>
     <DotNetCliVersion>2.0.2</DotNetCliVersion>
-    <VSWhereVersion>2.1.4</VSWhereVersion>
+    <VSWhereVersion>2.3.2</VSWhereVersion>
```
jaredpar: Why did this need to change?

build/build.ps1
```diff
@@ -21,7 +21,8 @@ Param(
 
 set-strictmode -version 2.0
 $ErrorActionPreference = "Stop"
-
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```
jaredpar: Is there a doc link that explains the implications of this change?

build/Versions.props
tmat: No particular reason, just getting the latest.

build/build.ps1
tmat: Other than https://msdn.microsoft.com/en-us/library/system.net.servicepointmanager.securityprotocol(v=vs.110).aspx?

PR #170, opened 2018-02-26
[WIP] Support collecting and reporting code coverage for tests

sharwell:
* Use OpenCover for collecting coverage data
* Report coverage data to codecov.io

⚠️ Please do not rebase or squash this pull request during a merge.


sharwell:
@tmat I derived this from the work in dotnet/roslyn-analyzers#1616 but I have no real way to test it that I see. Is there a way to create a roslyn-analyzers PR that uses a *PR* build of RepoToolset, as opposed to a build that got published?

JoeRobich:
@sharwell Is this something we should be considering for roslyn-tools or would it be a better fit for Arcade?

sharwell:
Probably arcade, but not sure

PR #179, opened 2018-03-02
Share obj directory between configurations

tmat:
Changes IntermediateOutputPath so that the `obj` directory of each project is shared between its Release and Debug builds. This allows NuGet to restore once and the generated NuGet files applied to both configurations.

Includes platform (x86, x64) in output paths.

### DIFF
--- a/sdks/RepoToolset/tools/ProjectLayout.props
+++ b/sdks/RepoToolset/tools/ProjectLayout.props
@@ -13,10 +13,14 @@
 
   <PropertyGroup>
     <OutDirName Condition="'$(OutDirName)' == ''">$(MSBuildProjectName)</OutDirName>
+
     <BaseOutputPath Condition="'$(BaseOutputPath)' == ''">$([System.IO.Path]::GetFullPath('$(ArtifactsBinDir)$(OutDirName)\'))</BaseOutputPath>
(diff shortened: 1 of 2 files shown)


tmat:
@tannergooding @jaredpar @nguerrera

tannergooding:
Doesn't this mess with incremental build/etc?

tmat:
@dsplaisted 
I guess we also need to add `$(Platform)` to address this:
https://github.com/Microsoft/msbuild/blob/b19a0906d1d20652f4ddb5fa9ce054b8692642f3/src/Directory.BeforeCommon.targets#L7-L14

tmat:
@tannergooding Why would it?

tmat:
@KevinH-MS Is this layout is ok for LUT?

tannergooding:
Ah, I missed that this is causing `BaseIntermediate` to be shared but also changes `Intermediate` to still contain the configuration

tmat:
I have added one more commit that implements support for multiple target Platforms.

tmat:
@dsplaisted This should allow msbuild to avoid setting the IntermediateOutputPath themselves.

sharwell:
@tmat can you update the PR top comment to include an overview of the behavior change?

sharwell:
@tmat that is an outstanding description :100:

PR #188, opened 2018-03-16
Multi proj build

tmat:


sdks/RepoToolset/tools/Build.proj
```diff
@@ -29,15 +30,23 @@
...
+  </ItemGroup>
+
+  <PropertyGroup>
+    <_ProjectsToBuild>@(_ProjectToBuild, ';')</_ProjectsToBuild>
+  </PropertyGroup>
+
   <PropertyGroup>
     <RepoRoot Condition="'$(RepoRoot)' != ''">$([System.IO.Path]::GetFullPath('$(RepoRoot)\'))</RepoRoot>
-    <RepoRoot Condition="'$(RepoRoot)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(SolutionPath), 'global.json'))\</RepoRoot>
+    <RepoRoot Condition="'$(RepoRoot)' == ''">$([MSBuild]::GetDirectoryNameOfFileAbove($(_ProjectsToBuild.Split(';')[0]), 'global.json'))\</RepoRoot>
```
tmat: @nguerrera :-|

tmat:
@weshaggard FYI - you might find this interesting.

jaredpar:
What was the motivation for adding this support?

Not against it. It's a scenario I've struggled with in Roslyn: do we merge everything into one happy solution or understand long term that we have a multi-solution repository. Wanted to see other motivations.

sdks/RepoToolset/tools/Build.proj
```diff
@@ -3,11 +3,12 @@
   <!--
   
   Required parameters:
-    SolutionPath                    Path to the solution to build.
+    Projects                        List of projects to build. Semicolon separated, may include globs.
```
jaredpar: Must this be a list of projects or can it be a list of solutions as well?

sdks/RepoToolset/tools/Build.proj
```diff
@@ -29,15 +30,23 @@
   <!--
     Default values. 
   -->
+  <ItemGroup>
+    <_ProjectToBuild Include="$(Projects)"/>
```
weshaggard: Where is the correct place for a repo using this toolset to define these Projects?

sdks/RepoToolset/tools/Build.proj
tmat: Both solutions and projects can be listed.

sdks/RepoToolset/tools/Build.proj
tmat: It's passed in as parameter from `build.ps1`. `build.ps1` takes an optional `-solution` parameter that can specify it. If the parameter is not passed we use `RepoRoot\*.sln`.

tmat: It would be probably good to have some kind of mechanism how to specify these in a file somewhere in the repo. I'm open to proposals how to do this. Most repos on RepoToolset have a single solution in the repo root.

tmat:
One motivation is to be able to avoid hardcoding solution name in `build.ps1` and `build.sh`. So these now can specify `$RepoRoot\*.sln` if the repo contains a single solution in the root (most do). 
Another motivation was @weshaggard asking for it for CoreFX style repos that have a bunch of separate solutions :)

weshaggard:
>One motivation is to be able to avoid hardcoding solution name in build.ps1 and build.sh. So these now can specify $RepoRoot\*.sln if the repo contains a single solution in the root (most do).
Another motivation was @weshaggard asking for it for CoreFX style repos that have a bunch of separate solutions.

To be clear for corefx I suspect we will use a traversal proj instead of passing in the multiple solutions.  We currently only use the solutions for working in VS, that may change but for now that is all we use them for.

sdks/RepoToolset/tools/Build.proj
weshaggard: I think we could set a property in the root Directory.Build.props file that identifies the root project(s).

tmat: That's too late. Directory.Build.props are imported way too late. It's much easier to check for an optional file of well-known name in the repo root. For example, `Solutions.props` or something like that. 
Have you looked into https://github.com/Microsoft/MSBuildSdks/tree/master/src/Traversal? Could that work?

weshaggard: >That's too late. Directory.Build.props are imported way too late. It's much easier to check for an optional file of well-known name in the repo root. For example, Projects.props or something like that.

OK I will take a deeper look at the ordering to see if there is a good place for it. 

> Have you looked into https://github.com/Microsoft/MSBuildSdks/tree/master/src/Traversal? Could that work?

Not in any detail yet but it is on my list.

PR #190, opened 2018-03-17
Add support for Reference Assemblies packages

tmat:
Implement https://github.com/dotnet/designs/pull/33 in RepoToolset.

The package that's referenced pulls down only the reference assemblies for target frameworks that the project targets.

### DIFF
--- a/sdks/RepoToolset/tools/ProjectDefaults.props
+++ b/sdks/RepoToolset/tools/ProjectDefaults.props
@@ -132,4 +132,8 @@
     </When>
   </Choose>
 
+  <ItemGroup Condition="'$(UsingToolNetFrameworkReferenceAssemblies)' == 'true'">
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="$(MicrosoftNetFrameworkReferenceAssembliesVersion)" PrivateAssets="all"/>
+  </ItemGroup>
+
 </Project>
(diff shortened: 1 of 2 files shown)


tmat:
@dsplaisted PTAL
@jaredpar @weshaggard @chcosta @markwilkie @natemcmaster FYI

sdks/RepoToolset/tools/ProjectDefaults.props
```diff
@@ -132,4 +132,11 @@
     </When>
   </Choose>
 
+  <!-- 
+    Implements proposal https://github.com/dotnet/designs/pull/33.
+  -->
+  <ItemGroup Condition="'$(UsingToolNetFrameworkReferenceAssemblies)' == 'true'">
```
dsplaisted: I'd suggest also conditioning this on whether the `TargetFrameworkIdentifier` is `.NETFramework`.

tmat: Wouldn't they be ignored if not? The build works with just this condition. Is there a potential problem?

dsplaisted: It's not a functional issue, but you'll end up with package references that aren't needed for projects that target something besides .NET Framework.  They won't actually do anything due to the logic inside them, but it seems cleaner not to reference them when you know they won't be needed.

PR #198, opened 2018-03-19
Remove unused line

chcosta:
This line led me to believe that reporoot would default to the directory of `global.json`, but it appears that auto-discovery was intentionally disabled via https://github.com/dotnet/roslyn-tools/commit/23bab92f5e8242e6a4f564cc94cb943c29f88e50

At this point, RepoRoot has already been set to `$([System.IO.Path]::GetFullPath('$(RepoRoot)/'))` which corresponds to the current directory because `/` gets resolved.

@tmat

### DIFF
--- a/sdks/RepoToolset/tools/RepoLayout.props
+++ b/sdks/RepoToolset/tools/RepoLayout.props
@@ -48,8 +48,6 @@
   </PropertyGroup>
   
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


tmat:
👎 This line is necessary for building from VS.

tmat:
I'll add a check that the RepoRoot contains global.json, to enforce consistency between CMD line build and VS build.

I removed the autodetection from build.props since I got some odd msbuild behavior

PR #204, opened 2018-03-21
Allow projects to opt-out of source build

tmat:
Implements `ExcludeFromSourceBuild` property that projects can set to exclude themselves from source build.

### DIFF
--- a/sdks/RepoToolset/sdk/Sdk.targets
+++ b/sdks/RepoToolset/sdk/Sdk.targets
@@ -4,6 +4,12 @@
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildThisFileFullPath);$(MSBuildAllProjects)</MSBuildAllProjects>
   </PropertyGroup>
+
+  <PropertyGroup>
+    <_SuppressAllTargets>false</_SuppressAllTargets>
+    <_SuppressAllTargets Condition="'$(DotNetBuildFromSource)' == 'true' and '$(ExcludeFromSourceBuild)' == 'true'">true</_SuppressAllTargets>
+  </PropertyGroup>
   
-  <Import Project="..\tools\Imports.targets" Condition="'$(__ExcludeSdkImports)' != 'true'" />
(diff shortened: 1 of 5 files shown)


tmat:
@nguerrera @dsplaisted @weshaggard @jaredpar This is slightly hacky but I think it's better than maintaining extra source-build-only solutions or adding conditions to each project that multi-targets. 

It would be nice if NuGet Restore had more supported way to skip restore of a project from within the project. I could only find `_IsProjectRestoreSupported` target that does what we need here.

sdks/RepoToolset/sdk/Sdk.targets
```diff
@@ -4,6 +4,35 @@
...
+  -->
+
+  <PropertyGroup>
+    <_SuppressBuild>false</_SuppressBuild>
+    <_SuppressBuild Condition="'$(__ExcludeSdkImports)' == 'true'">true</_SuppressBuild>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(DotNetBuildFromSource)' == 'true' and !$(_SuppressBuild)">
+    <_ReducedTargetFrameworks Condition="'$(TargetFrameworks)' != ''">$([System.Text.RegularExpressions.Regex]::Replace('$(TargetFrameworks)', 'net[0-9]+;?', '').Trim(';'))</_ReducedTargetFrameworks>
+    <_ReducedTargetFrameworks Condition="'$(TargetFrameworks)' == ''">$([System.Text.RegularExpressions.Regex]::Replace('$(TargetFramework)', 'net[0-9]+', ''))</_ReducedTargetFrameworks>
```
jaredpar: We want to keep all the net* targets though. RedHat was okay with us doing that as we're just compiling against the equivalent of header files here.

tmat: I see. The projects that reference non-FX non-OSS assemblies need to be filtered out (anything that builds against VS assemblies, for example), right?

tmat: Updated the filter to check for `ExcludeFromSourceBuild`. Projects can now set this property to exclude themselves from source build.

tmat:
test ci please

tmat:
test this please

PR #210, opened 2018-03-26
Report insertion error if latest successful build has already been inserted, but there are more recent unsuccessful builds

### LINKED ISSUE #206: Insertion tool should warn if it's skipping over newer builds
The insertion tool's logic is:

1. Find the oldest passing build.
2. Insert that.
3. If it turns out no files changed, then simply don't create a PR and call it a "skip".

This can backfire if the oldest passing build is the build that's already inserted, and the reason it picked that build is because the newer builds were all failing. If that's the case, it should treat that as failure, not as a false "success".

dsplaisted:
Fixes #206

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -86,9 +86,24 @@ namespace Roslyn.Insertion
             var definitions = await buildClient.GetDefinitionsAsync(project: Options.TFSProjectName, name: Options.BuildQueueName);
             var builds = await GetBuildsFromTFSAsync(buildClient, definitions, cancellationToken, resultFilter: null);
 
-            return (from build in builds
+            return (await BuildsWithValidArtifactsAsync(buildClient, cancellationToken, from build in builds
                     orderby build.FinishTime descending
-                    select build).FirstOrDefault();
+                    select build)).FirstOrDefault();
+        }
+
+        private static async Task<List<Build>> BuildsWithValidArtifactsAsync(BuildHttpClient buildClient, CancellationToken cancellationToken,
+                        IEnumerable<Build> builds)
+        {
+            List<Build> buildsWithValidArtifacts = new List<Build>();
+            foreach (var build in builds)
+            {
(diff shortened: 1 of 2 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -96,14 +81,34 @@ where buildClient.GetArtifactsAsync(build.Project.Id, build.Id, cancellationToke
...
             {
-                newestBuild = await GetLatestPassedBuildAsync(cancellationToken);
+                Log.Trace($"Getting latest passing build from {Options.TFSProjectName} where name is {Options.BuildQueueName}");
+                var buildClient = ProjectCollection.GetClient<BuildHttpClient>();
+                var definitions = await buildClient.GetDefinitionsAsync(project: Options.TFSProjectName, name: Options.BuildQueueName);
+                var builds = await GetBuildsFromTFSAsync(buildClient, definitions, cancellationToken, BuildResult.Succeeded);
+
+                // Get the latest build with valid artifacts.
+                newestBuild = (from build in builds
+                        where buildClient.GetArtifactsAsync(build.Project.Id, build.Id, cancellationToken).Result.Any(a => a.Resource != null && a.Resource.Data != null && a.Resource.Data.Contains(Options.BuildDropPath))
```
jasonmalinowski: Note we still need this check in for the "latest" build -- this filters out prodcon builds that we can't insert either.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -65,20 +66,26 @@ public static partial class RoslynInsertionTool
                 cancellationToken.ThrowIfCancellationRequested();
 
                 BuildVersion buildVersion;
-                Build newestBuild;
+
+                Build buildToInsert;
```
jasonmalinowski: I approve of this rename.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
dsplaisted: @jasonmalinowski I think I've fixed this.  Unfortunately LINQ doesn't mix well with async so it's uglier now, but we're also not accessing the `Result` of an async method's return value anymore either.

PR #211, opened 2018-03-28
Source Build fixes

tmat:
A few of fixes for source build and Linux build:
- Respect `DotNetRestoreSourcePropsPath` and `DotNetBuildOffline` source build variables
- Use `Microsoft.NETCore.Compilers` package when building via dotnet CLI on CI.
- Use `Microsoft.NET.Compilers` package when building via desktop msbuild on CI.
- Use locally installed compilers when building on dev machine.

### DIFF
--- a/sdks/RepoToolset/tools/DefaultVersions.props
+++ b/sdks/RepoToolset/tools/DefaultVersions.props
@@ -58,8 +58,8 @@
   <!-- RestoreSources overrides - defines DotNetRestoreSources variable if available -->
   <Import Project="$(DotNetRestoreSourcePropsPath)" Condition="'$(DotNetRestoreSourcePropsPath)' != ''"/>
 
-  <!-- Force sources to DotNetRestoreSources if specified, ignore any sources set by the repo -->
-  <PropertyGroup Condition="'$(DotNetRestoreSources)' != ''">
+  <!-- Force sources to DotNetRestoreSources if building offline, discard any sources set by the repo -->
+  <PropertyGroup Condition="'$(DotNetBuildOffline)' == 'true'">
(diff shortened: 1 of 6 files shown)


tmat:
@tannergooding @nguerrera @jaredpar @dseefeld PTAL

sdks/RepoToolset/tools/DefaultVersions.props
```diff
@@ -49,10 +55,18 @@
   <Import Project="$(DotNetPackageVersionPropsPath)" Condition="'$(DotNetPackageVersionPropsPath)' != ''" />
   <Import Project="$(FixedVersionsPropsPath)" Condition="'$(FixedVersionsPropsPath)' != ''"/>
 
-  <PropertyGroup>
+  <!-- RestoreSources overrides - defines DotNetRestoreSources variable if available -->
+  <Import Project="$(DotNetRestoreSourcePropsPath)" Condition="'$(DotNetRestoreSourcePropsPath)' != ''"/>
+
+  <!-- Force sources to DotNetRestoreSources if specified, ignore any sources set by the repo -->
+  <PropertyGroup Condition="'$(DotNetRestoreSources)' != ''">
+    <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
```
dseefeld: If there are sources set by the repo, DotNetRestoreSources shouldn't always override them.  If we're building online, DotNetRestoreSources should be the first, but others need to be present to allow repos to find their packages in the online build.  For offline build, DotNetRestoreSources should be the only one.

tmat: OK, I'll add the offline check.

PR #226, opened 2018-04-02
Option to have an MSBuild binary log emitted

jaredpar:
This is necessary with new signing to get diagnostic output about why
the signing is failing.


jaredpar:
@tmat the linux legs are failing even though everything succeeded. Are we still ignoring failures in those legs?

tmat:
@jaredpar Yes.

PR #237, opened 2018-04-11
Update XliffTasks version

tmeschter:
Update the version of XliffTasks used by the RepoToolset. Notably, this
new version inserts new `trans-unit` elements into .xlf files in sort
order rather than always putting them at the end. This should help
reduce merge conflicts when the .xlf files have changed in multiple
places.


tmat:
👍

PR #238, opened 2018-04-13
Update artifact name

jasonmalinowski:
Oops, I renamed this.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -213,7 +213,7 @@ namespace Roslyn.Insertion
 
             var artifactMetadata = new ArtifactMetadata
             {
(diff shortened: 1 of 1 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -213,7 +213,7 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
 
             var artifactMetadata = new ArtifactMetadata
             {
-                Alias = "Roslyn-Signed", // This is the alias in the Release definition.
+                Alias = "InputBuild", // This is the alias in the Release definition.
```
brettfo: Do other build definitions need to specify an alias?  What's the benefit?

jasonmalinowski: @brettfo We use the same release queue for all definitions. When it was setup the alias was called "Roslyn-Signed" because the default alias name is just the name of the build definition it's connected to. I renamed it to InputBuild so it's clearer it's just any random build.

PR #241, opened 2018-04-20
Update insertion tool

brettfo:
Update the insertion tool to handle new values being added to versions.xml and AssemblyVersions.tt.

I ran these changes locally and I can show you what the updated insertion looks like.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/VersionsUpdater.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/VersionsUpdater.cs
@@ -148,7 +148,7 @@ namespace Roslyn.Insertion
 
             var attributes = (from element in sourceElements
                               where element.Attribute("name")?.Value == variableName
-                              select element.Attribute("value"))?.ToArray();
+                              select element.Attribute("value")).ToArray();
 
             if (attributes.Length == 0)
             {
@@ -223,12 +223,12 @@ namespace Roslyn.Insertion
                 if (comparison == 0)
                 {
                     // found exact match, replace this line
-                    var versionStart = line.IndexOf('"') + 1;
-                    var versionEnd = line.IndexOf('"', versionStart);
+                    var versionStart = IndexOfOrThrow(line, '"') + 1;
+                    var versionEnd = IndexOfOrThrow(line, '"', versionStart);
                     var versionStr = line.Substring(versionStart, versionEnd - versionStart);
                     ParseAndValidatePreviousVersion(newVersion, versionStr, fullPath, variableName, assemblyName);
-                    var existingLine = line.Substring(0, versionStart) + newVersion.ToFullVersion() + line.Substring(versionEnd);
-                    sortedValueLines[lineIndex] = existingLine;
+                    newLine = line.Substring(0, versionStart) + newVersion.ToFullVersion() + line.Substring(versionEnd);
+                    sortedValueLines[lineIndex] = newLine;
                     valueUpdated = true;
                     break;
                 }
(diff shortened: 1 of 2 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/VersionsUpdater.cs
```diff
@@ -169,24 +188,67 @@ private void UpdateAssemblyVersionsFile(string assemblyName, Version newVersion)
...
+                var comparison = String.Compare(currentVariableName, variableName, StringComparison.OrdinalIgnoreCase);
+                if (comparison == 0)
+                {
+                    // found exact match, replace this line
+                    var versionStart = line.IndexOf('"') + 1;
+                    var versionEnd = line.IndexOf('"', versionStart);
+                    var versionStr = line.Substring(versionStart, versionEnd - versionStart);
+                    ParseAndValidatePreviousVersion(newVersion, versionStr, fullPath, variableName, assemblyName);
+                    var existingLine = line.Substring(0, versionStart) + newVersion.ToFullVersion() + line.Substring(versionEnd);
+                    sortedValueLines[lineIndex] = existingLine;
```
jasonmalinowski: Should we just set this to newLine rather than trying to preserve this?

src/RoslynInsertionTool/RoslynInsertionTool/VersionsUpdater.cs
```diff
@@ -169,24 +188,67 @@ private void UpdateAssemblyVersionsFile(string assemblyName, Version newVersion)
...
+            for (int lineIndex = 0; lineIndex < sortedValueLines.Count; lineIndex++)
             {
-                throw new InvalidDataException($"File '{fullPath}' doesn't have expected format.");
+                var line = sortedValueLines[lineIndex];
+                var currentVariableName = GetLineVariableName(line);
+                var comparison = String.Compare(currentVariableName, variableName, StringComparison.OrdinalIgnoreCase);
+                if (comparison == 0)
+                {
+                    // found exact match, replace this line
+                    var versionStart = line.IndexOf('"') + 1;
```
jasonmalinowski: BTW: all these .IndexOfs that will return -1 if not found are terrifying me. Wonder if we should have ThrowingIndexOf or something.

src/RoslynInsertionTool/RoslynInsertionTool/VersionsUpdater.cs
```diff
@@ -119,36 +119,55 @@ private void UpdateVersionsFile(string assemblyName, Version newVersion)
...
-                throw new InvalidDataException($"Missing version element with @name='{variableName}' (file '{fullPath}')");
+                throw new InvalidDataException($"Invalid XML data structure.  Expected: /root/versions/version (file '{fullPath}')");
             }
 
-            var valueAttribute = attributes.Single();
-
-            ParseAndValidatePreviousVersion(newVersion, valueAttribute.Value, fullPath, "version.@name", assemblyName);
+            var attributes = (from element in sourceElements
+                              where element.Attribute("name")?.Value == variableName
+                              select element.Attribute("value"))?.ToArray();
```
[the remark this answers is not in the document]
brettfo: Thanks, I meant to remove that but forgot.

src/RoslynInsertionTool/RoslynInsertionTool/VersionsUpdater.cs
```diff
@@ -169,24 +188,67 @@ private void UpdateAssemblyVersionsFile(string assemblyName, Version newVersion)
...
             }
 
-            var i = content.IndexOf(variableName);
-            if (i < 0)
+            // the structure of the file is:
+            //   header: opening tag:    <#+
+            //           comments:           //
+            //   values:                     const string AssemblyNameVersion = "<version-number>";
+            //   footer: closing tag:    #>
+            var lines = content.Split('\n').Select(line => line.TrimEnd('\r')).ToList();
```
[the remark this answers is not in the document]
brettfo: I did this on purpose on the off chance that the local git install is ever updated to check out as `\n` instead of `\r\n`; this way it always works (TM).

src/RoslynInsertionTool/RoslynInsertionTool/VersionsUpdater.cs
brettfo: Good call, I'll do that.

src/RoslynInsertionTool/RoslynInsertionTool/VersionsUpdater.cs
brettfo: Done.

PR #247, opened 2018-04-30
Update XliffTasks version

tmeschter:
Update to the latest version of XliffTasks. This version includes fixes
to make a couple of targets (for sorting .xlf files and verifying all
resources are localized) work properly on projects with multiple target
frameworks.


tmeschter:
Superceded by #266.

PR #250, opened 2018-05-02
Compiler preview installer

jmarolf:
Following features have been added:
- [X] User can choose which version of VS preview it installs into
- [x] Only installs on VS versions 15.5-15.7
- [X] Clear out MEF Cache on install/uninstall to fix reliability issues
- [X] Attempt uninstall multiple times on error as it usually succeeds on subsequent re-tries
- [x] Autmatically handles case of older preview being installed

Not Tested
- [ ] Installation on Visual Studio versions 15.5 and 15.6

Not Implemented
- [ ] Installation of Range feature branch (should be drop-in though)
- [ ] User can install multiple previews into hives


installers/Roslyn_Nullable_References_Preview/tools/install.ps1
```diff
@@ -1,50 +1,68 @@
...
 
-    Deploy-VsixViaTool
+    # Welcome Message
+    Write-Host "Installing C# 8.0 Nullable Preview" -ForegroundColor Green
+
+    # Find VS Instance
+    $vsInstalls = Get-VisualStudioDirAndId
+    $vsDir = $vsInstalls[0].Trim("\")
+    $vsId = $vsInstalls[1]
+    if ($vsInstalls.Count -ge 3) {
```
jaredpar: Why 3 here and not 2?

installers/Roslyn_Nullable_References_Preview/tools/install.ps1
```diff
@@ -1,50 +1,68 @@
...
+          $vsId = $vsInstalls[$index+1]
+          break
+        }
+
+        Write-Host ""
+      }
+    }
+
+    # Install VSIX
+    Write-Host "Installing Preview..." -ForegroundColor Gray
```
jaredpar: Consider adding the path to the Visual Studio here that the preview is being installed into.

installers/Roslyn_Nullable_References_Preview/tools/install.ps1
```diff
@@ -1,50 +1,68 @@
...
+    Uninstall-VsixViaTool -vsDir $vsDir -vsId $vsId -hive ""
+    Uninstall-OlderVsixesViaTool -vsDir $vsDir -vsId $vsId -hive ""
+    Install-VsixViaTool -vsDir $vsDir -vsId $vsId -hive ""
+
+    # Clear MEF Cache
+    $mefCacheFolder = Join-Path $env:LOCALAPPDATA "Microsoft\VisualStudio\15.0_$vsId\ComponentModelCache"
+    Get-ChildItem -Path $mefCacheFolder -Include *.* -File -Recurse | foreach { $_.Delete()}
```
jaredpar: Is there an advantage to `$_.Delete` vs. `Remove-Item $_`?

installers/Roslyn_Nullable_References_Preview/tools/install.ps1
```diff
@@ -1,50 +1,69 @@
...
+    Exec-Console $vsExe $args
+
+    Write-Host "Install Succeeded" -ForegroundColor Green
     exit 0
 }
 catch {
-    Write-Host $_
-    Write-Host $_.Exception
-    Write-Host $_.ScriptStackTrace
+    Write-Host $_ -ForegroundColor Red
```
tannergooding: Would `Write-Error` be better here?

tannergooding: and the other places where we are printing error messages?

installers/Roslyn_Nullable_References_Preview/tools/uninstall.ps1
```diff
@@ -1,49 +1,43 @@
...
     if (Test-Process "devenv") {
-        Write-Host "Please shut down all instances of Visual Studio before running"
+        Write-Host "Please shut down all instances of Visual Studio before running" -ForegroundColor Red
         exit 1
     }
 
-    Uninstall-VsixViaTool
+    # Find VS Instance
+    $vsInstalls = Get-VisualStudioDirAndId
+    $vsDir = $vsInstalls[0].Trim("\")
```
tannergooding: Why trim the trailing slash?

installers/Roslyn_Nullable_References_Preview/tools/install.ps1
jaredpar: No. Write-Error adds extra infromation to the output that obscures what we're trynig to print.

installers/Roslyn_Nullable_References_Preview/tools/uninstall.ps1
jmarolf: vsexpinstaller doesn't like it

installers/Roslyn_Nullable_References_Preview/tools/utils.ps1
```diff
@@ -76,33 +76,105 @@ function Exec-Console([string]$command, [string]$commandArgs) {
...
+    Write-Host $ErrorInfo.Exception  -ForegroundColor Red
+    Write-Host $ErrorInfo.ScriptStackTrace -ForegroundColor Red
+    exit 1
+  }
+}
+
+function Uninstall-OlderVsixesViaTool([string]$vsDir, [string]$vsId, [string]$hive) {
+  $baseArgs = "/rootSuffix:$hive /u /vsInstallDir:`"$vsDir`""
+
+  $all = @(
```
tannergooding: Is it worth noting that these need to remain ordered, due to the dependency chain?

tannergooding:
Do we need to check the vsix in? Can we not put it in blob storage or elsewhere, to prevent the repo-bloat

installers/Roslyn_Nullable_References_Preview/tools/uninstall.ps1
tannergooding: Is it interpreting the `"pathwithtrailingslash"` as having unmatching quotes due to the `\"`?

installers/Roslyn_Nullable_References_Preview/tools/utils.ps1
```diff
@@ -76,33 +76,105 @@ function Exec-Console([string]$command, [string]$commandArgs) {
...
+    "vsix\ExpressionEvaluatorPackage.vsix"
+    "vsix\Roslyn.VisualStudio.InteractiveComponents.vsix",
+    "vsix\Roslyn.VisualStudio.Setup.Next.vsix",
+    "vsix\Roslyn.VisualStudio.Setup.vsix",
+    "vsix\Roslyn.Compilers.Extension.vsix")
+
+  $attempt = 6
```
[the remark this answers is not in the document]
jmarolf: Surprisingly yes, I have had situations where each individual install will fail, but subsequently succeed if immediately retried. We install 5 vsix so 5+1 = 6

installers/Roslyn_Nullable_References_Preview/tools/utils.ps1
jmarolf: If there was an install step for them maybe, but these are only here to handle an edge case.

installers/Roslyn_Nullable_References_Preview/tools/utils.ps1
```diff
@@ -76,33 +76,155 @@ function Exec-Console([string]$command, [string]$commandArgs) {
...
-
-        # Need to be using at least Visual Studio 15.2 in order to have the appropriate
+        # Need to be using at least Visual Studio 15.5 in order to have the appropriate
         # set of SDK fixes. Parsing the installationName is the only place where this is
         # recorded in that form.
         $name = $obj.installationName
         if ($name -match "VisualStudio(Preview)?/([\d.]+)(\+|-).*") {
-            $minVersion = New-Object System.Version "15.3.0"
+            $minVersion = New-Object System.Version "15.5.0"
+            $maxVersion = New-Object System.Version "15.7.0"
```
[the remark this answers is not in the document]
jmarolf: Yeah, I should change this to be "up-to but not inclusive of 15.8"

jcouv:
@jmarolf Given the issues discussed above (version range, hive option), would it be reasonable to aim for making a new installer early next week? We have a few compiler fixes that we'll resolve in that same timeframe.

Tagging @cston as FYI

jmarolf:
@jcouv I am going to fix the version range issue right now.  Whenever there is an update we can rev the installer.  We shouldn't need to wait for a specific set of fixes to flow in.

jcouv:
@jmarolf Ok. If that's the case, let's go ahead with just the installer changes (thanks!). 
We'll do the compiler fixes separately next week.

installers/Roslyn_Nullable_References_Preview/tools/uninstall.ps1
```diff
@@ -1,49 +1,44 @@
...
+    $vsId = $vsInstalls[1]
+
+    # Uninstall VSIX
+    Write-Host "Uninstallting Preview Everywhere..." -ForegroundColor Green
```
cston: Typo: `"Uninstalling ..."`

jmarolf:
superseded by: https://github.com/dotnet/roslyn/pull/26853

PR #251, opened 2018-05-02
Use drop path from API for insertions

agocke:
We were previously using a hardcoded path to find the drop path for
official builds. This changes the insertion tool to use the artifact
build pulled directly out of the VSTS API.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -191,6 +191,32 @@ namespace Roslyn.Insertion
                     select build).FirstOrDefault();
         }
 
+        internal static async Task<string> GetBuildDirectoryAsync(Build build, CancellationToken cancellationToken)
+        {
+            // used for local testing:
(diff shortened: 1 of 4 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -191,6 +191,32 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
...
+            var buildClient = ProjectCollection.GetClient<BuildHttpClient>();
+
+            // Pull the build directory from the API
+            var artifacts = await buildClient.GetArtifactsAsync(build.Project.Id, build.Id, cancellationToken);
+            foreach (var artifact in artifacts)
+            {
+                // The drop path we're looking for is named the same as the build number
+                if (artifact.Name == build.BuildNumber)
+                {
+                    return Path.Combine(artifact.Resource.Data, build.BuildNumber);
```
brettfo: Will you add a note here that `artifact.Resource.Data` will always have a value because of line 101 above?  I can see myself changing line 101 and wondering why this part might fail.

agocke: Absolutely

PR #253, opened 2018-05-07
Add property to control whether RoslynTools Sdk imports Microsoft.NET.Sdk

### LINKED ISSUE #249: WorkaroundIncorrectSatelliteDeployment target can probably be removed from workarounds
The `WorkaroundIncorrectSatelliteDeployment` has a comment on it:

> Work around bug in Microsoft.NET.Sdk < v2.0 where satellites were deployed on top of each other in root. https://github.com/dotnet/sdk/issues/1360

The SDK bug has been fixed, so this target can probably be removed.  The workaround is causing issues when trying to pack a project:

> C:\git\dotnet-sdk\artifacts\.dotnet\2.1.300-preview3-008616\sdk\2.1.300-preview3-008616\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(198,5): error NU5118: File 'C:\git\dotnet-sdk\artifacts\Debug\bin\Tests\Microsoft.NET.Build.Tests\publish\de\Microsoft.DotNet.Cli.Utils.resources.dll' is not added because the package already contains file 'tools\netcoreapp2.1\any\de\Microsoft.DotNet.Cli.Utils.resources.dll' [C:\git\dotnet-sdk\src\Tests\Microsoft.NET.Build.Tests\Microsoft.NET.Build.Tests.csproj]

dsplaisted:
This adds an `ImportNetSdkFromRepoToolset` property which can be used to control whether the RoslynTools.RepoToolset Sdk imports Microsoft.NET.Sdk.  It defaults to true (so behavior is unchanged), but if you set it to false, you can use Microsoft.NET.Sdk (or Microsoft.NET.Sdk.Web) in your project files.  This means that new project templates will work without modification.

In order to do this, you can put the following in a Directory.Build.props file:

```xml
  <PropertyGroup>
    <ImportNetSdkFromRepoToolset>false</ImportNetSdkFromRepoToolset>
  </PropertyGroup>

  <Import Project="Sdk.props" Sdk="RoslynTools.RepoToolset" />
```

Similarly, you would put the following in Directory.Build.targets:

```xml
<Import Project="Sdk.targets" Sdk="RoslynTools.RepoToolset" />
```

Additionally, this PR removes the `WorkaroundIncorrectSatelliteDeployment` target, which fixes #249

### DIFF
--- a/sdks/RepoToolset/tools/Imports.targets
+++ b/sdks/RepoToolset/tools/Imports.targets
@@ -6,7 +6,7 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
 
-  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
+  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" Condition="'$(ImportNetSdkFromRepoToolset)' == 'true'" />
   
   <PropertyGroup>
     <DeployProjectOutput Condition="'$(DeployProjectOutput)' == ''">$(__DeployProjectOutput)</DeployProjectOutput>
--- a/sdks/RepoToolset/tools/Settings.props
+++ b/sdks/RepoToolset/tools/Settings.props
@@ -4,6 +4,8 @@
 
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+
+    <ImportNetSdkFromRepoToolset Condition="'$(ImportNetSdkFromRepoToolset)' == ''">true</ImportNetSdkFromRepoToolset>
   </PropertyGroup>
 
   <Import Project="BuildTasks.props" />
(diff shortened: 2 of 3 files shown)


tmat:
The problem with this change is that if you import RepoToolset in Directory.Build.props the Settings.props will be imported after Microsoft.Net.Sdk, not before as it is today. 
The toolset currently makes some assumptions on the ordering. Setting ImportNetSdkFromRepoToolset will break those assumptions.

nguerrera:
Directory.Build.props is before the SDK or Common.props doe anything meaningful.

tmat:
Yes, but wouldn't the project import Microsoft.Net.Sdk sdk.props first (because it's in `<Project Sdk="Microsoft.Net.Sdk">`) before Directory.Build.props imports RoslynTools.RepoToolset sdk.props?

nguerrera:
Technically, yes, but it will go like this:

```
Import sdk.props
    (Nothing of significance in sdk.props)
    Import Common.props
        (Nothing of significance in Common.props)
        Import Directory.Build.props
            Import repotoolset.props
       (Interesting parts of common.props)
    (Interesting parts of sdk.props)
```

Directory.Build.props is commonly used precisely to put things before the sdk and common props (even though it's not literally before). I would think this would work fine for repo toolset too.

tmat:
I guess that should work then. In some cases though we need the toolset to have complete control of what's imported.

See https://github.com/dotnet/roslyn-tools/blob/master/sdks/RepoToolset/sdk/Sdk.targets#L8-L15

Projects can specify `ExcludeFromSourceBuild` property to remove the project from the build.

dsplaisted:
@tmat Could you take a look at this when you get a chance?  Here's a commit for the SDK that shows how it would be used: https://github.com/dsplaisted/sdk/commit/4a37cee0adfa9f91002df445ceaed13e2d275b70

dsplaisted:
@tmat I tested building dotnet/sdk modified to take advantage of these changes with DotNetBuildFromSource set to true, and the overrides from Empty.targets in repotoolset still applied.

tmat:
Ok, Cool.

PR #257, opened 2018-05-10
Add automerge bot

agocke:
If a pull request is marked with the label "Auto-Merge If Tests Pass"
and this bot is enabled on the repo, the bot will automatically approve
and merge PRs when all merge requirements are satisfied.

### DIFF
--- a/src/GithubMergeTool/GithubMergeTool.cs
+++ b/src/GithubMergeTool/GithubMergeTool.cs
@@ -215,6 +215,15 @@ Once all conflicts are resolved and all the tests pass, you are free to merge th
                 return (false, null);
             }
 
+            Console.WriteLine(prInfo.ToString());
+
+            // Check that the PR is by 'dotnet-bot'. Eventually we will support created PRs
+            // created by other users, but not right now.
+            if ((string)prInfo["user"]["login"] != "dotnet-bot")
+            {
+                return (false, null);
+            }
+
             // Check that the PR has a well-known 'auto-merge' label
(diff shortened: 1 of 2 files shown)


src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -163,5 +164,158 @@ git commit
...
+                {
+                    // There is a failing required test
+                    return (false, null);
+                }
+            }
+
+            // Check if there are any approvals
+            if (!prReviews.Any(review => (string)review["state"] == "APPROVED"))
+            {
+                // If not, mark the PR as approved
```
brettfo: I would argue that approving a PR is beyond the scope of this merge tool:

1. I might want to add the appropriate label to a PR, but I could still be waiting on code review/approval.  E.g., an escrow change.
2.  The account that creates a PR can't be the one that approves it so this will regularly fail.  My suggestion for this is to create another bot _and_ another account that scans for and approves the auto-merge PRs.  That way the busy work of merging is the responsibility of bots, but human-created PRs can still benefit from the auto-merging, but only after getting a real person to approve it.

agocke: > I might want to add the appropriate label to a PR, but I could still be waiting on code review/approval

Then you should block it using required tests, like marking the PR `[WIP]`.

> The account that creates a PR can't be the one that approves it so this will regularly fail.

This bot never creates PRs, it only approves them. So it will never fail.

PR #263, opened 2018-05-23
Fix to update packages even if they are published to the artifacts server instead of \\cpvsbuild

AbhitejJohn:
Changes include the following:
1. Download an artifact if it is published to the server and use that to update VS.
2. Repurpose `insertcoreXTPackages` switch to only updates the packages in default.config and not require an update to `ProductData\ContractAssemblies.props` etc that .Net libraries require. That is handled now by an existing `updateCoreXTPackages` flag which was from what I understand originally intended for this very purpose.


AbhitejJohn:
Tagging @jasonmalinowski , @Shyam-Gupta  for a review.

src/RoslynInsertionTool/RoslynInsertionTool.Commandline/RoslynInsertionToolCommandline.cs
```diff
@@ -151,6 +151,11 @@ partial class RoslynInsertionToolCommandline
                 $"Defaults to \"{options.InsertWillowPackages}\".",
                 insertWillowPackages => options = options.WithInsertWillowPackages(bool.Parse(insertWillowPackages))
             },
+            {
+                "il=|insertLibraryPackages=",
```
jasonmalinowski: Can we redo these switches to something a bit clearer? Perhaps "update packages in CoreXT" and "add binding redirect information" to more clearly define the behavior of these two?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -98,7 +98,10 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
             foreach (var build in builds)
             {
                 var artifacts = await buildClient.GetArtifactsAsync(build.Project.Id, build.Id, cancellationToken);
-                if (artifacts.Any(a => a.Resource != null && a.Resource.Data != null && a.Resource.Data.Contains(Options.BuildDropPath)))
+                // A valid artifact would:
+                // 1. Have a resource
+                // 2. The resource would either point to a cpvsbuild path (or) be uploaded to the artifacts service with a build number label.
+                if (artifacts.Any(a => a.Resource != null && a.Resource.Data != null && (a.Resource.Data.Contains(Options.BuildDropPath) || a.Resource.Data.Contains(build.BuildNumber))))
```
jasonmalinowski: Can we make this multiple lines?

jasonmalinowski: Does this imply our artifact name has to be the build number? We actually *shouldn't* be doing that IIRC. I know we have repos configured that way, but it's actually broken. :smile:

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -222,6 +232,39 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
             throw new InvalidOperationException("Could not find drop path");
         }
 
+        private static async Task<string> DownloadArtifactsAsync(BuildHttpClient buildClient, Build build, BuildArtifact artifact, CancellationToken cancellationToken)
+        {
+            var tempDirectory = Path.Combine(Path.GetTempPath(), string.Concat(Options.InsertionName, Options.BranchName).Replace(" ", "_").Replace("/","_"));
+            if (Directory.Exists(tempDirectory))
+            {
+                // Be judicious and clean up old artifacts so we do not eat up memory on the scheduler machine.
+                // Sometime FileSystem APIs do not create a directory immediately after a deletion - Looks like there is a race.
```
jasonmalinowski: Suspicious comment is suspicious. I wouldn't be so quick to blame file system APIs...

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -222,6 +232,39 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
...
+                // So I'm just cleaning up the old directories contents in this case.
+                foreach (var dir in Directory.GetDirectories(tempDirectory)) Directory.Delete(dir, recursive: true);
+                foreach (var file in Directory.GetFiles(tempDirectory)) File.Delete(file);
+            }
+            else
+            {
+                Directory.CreateDirectory(tempDirectory);
```
tannergooding: Don't we want to create the directory after deleting above?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -222,6 +232,39 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
...
 
+        private static async Task<string> DownloadArtifactsAsync(BuildHttpClient buildClient, Build build, BuildArtifact artifact, CancellationToken cancellationToken)
+        {
+            var tempDirectory = Path.Combine(Path.GetTempPath(), string.Concat(Options.InsertionName, Options.BranchName).Replace(" ", "_").Replace("/","_"));
+            if (Directory.Exists(tempDirectory))
+            {
+                // Be judicious and clean up old artifacts so we do not eat up memory on the scheduler machine.
+                // Sometime FileSystem APIs do not create a directory immediately after a deletion - Looks like there is a race.
+                // So I'm just cleaning up the old directories contents in this case.
+                foreach (var dir in Directory.GetDirectories(tempDirectory)) Directory.Delete(dir, recursive: true);
```
tannergooding: Why not just delete the entire tempDirectory at once?

AbhitejJohn: Added a comment in code on why I had to do what I did.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
AbhitejJohn: [This](https://stackoverflow.com/questions/32593552/using-directory-delete-and-directory-createdirectory-to-overwrite-a-folder) is the only reference I found that sympathizes with this change. Ran into this twice when testing locally.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
AbhitejJohn: oh, derived that from other places in code. I'm open to change it though. One option is to tie this with `droppath` and pass in a `server` in there. Thoughts?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
AbhitejJohn: I couldn't find any documentation for why this happens though, just observations from [here](https://stackoverflow.com/questions/32593552/using-directory-delete-and-directory-createdirectory-to-overwrite-a-folder) which I ran into during testing. Would be happy to try other suggestions out though.

src/RoslynInsertionTool/RoslynInsertionTool.Commandline/RoslynInsertionToolCommandline.cs
AbhitejJohn: hmm, looks like there is a `UpdateCoreXTLibraries` switch that we could potentially reuse. Dont know if that has other implications to it. [Here](https://github.com/dotnet/roslyn-tools/blob/master/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs#L101) is where it is used today and its default is false. Is this set to true for roslyn and project system insertions?

AbhitejJohn:
@jasonmalinowski , @tannergooding : Can you guys take one more look at this please? I've removed the new flag that I've added.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -222,6 +236,39 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
             throw new InvalidOperationException("Could not find drop path");
         }
 
+        private static async Task<string> DownloadArtifactsAsync(BuildHttpClient buildClient, Build build, BuildArtifact artifact, CancellationToken cancellationToken)
+        {
+            var tempDirectory = Path.Combine(Path.GetTempPath(), string.Concat(Options.InsertionName, Options.BranchName).Replace(" ", "_").Replace("/","_"));
+            if (Directory.Exists(tempDirectory))
```
tannergooding: I still think this is suspicious.

Doing:
```
if (Directory.Exists(tempDirectory))
{
    Directory.Delete(path, recursive: true);
}
Directory.CreateDirectory(path);
```
is pretty standard practice.

and if there really is a race condition, we could add some basic retry logic or also cleanup after the insertion completes (rather than just at the start of the current insertion).
* If there is a race condition, that also means that these folders could (theoretically) not be deleted by the time we start this insertion (and that could put us in a bad state)

AbhitejJohn: hmm, maybe [this other github issue](https://github.com/dotnet/corefx/issues/15330) might help. Looks like we would need to upgrade to a later .Net framework to get this working(4.7.1 maybe). I could try that out, wonder if the scheduler machine has that installed though?

tannergooding: @jasonmalinowski, what are your thoughts here?

jasonmalinowski: I'm OK upgrading if we think there's an actual fix (but I didn't see an actual fix with that issue?)

In any case, it seems the simpler code would just be:

```c#
while (Directory.Exists(tempDirectory))
try
{
    Directory.Delete(tempDirectory, recursive: true)
}
catch (Exception e) {}
```

or the likes. No reason to invent your first round of recursion here.

AbhitejJohn: Fixed.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -222,6 +236,43 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
...
+            {
+                // Be judicious and clean up old artifacts so we do not eat up memory on the scheduler machine.
+                Directory.Delete(tempDirectory, recursive: true);
+                // Sometimes a creation of a directory races with deletion since at least in .net 4.6 deletion is not a blocking call.
+                // Hence explictly waiting for the directory to be deleted before moving on.
+                Stopwatch w = Stopwatch.StartNew();
+                while (Directory.Exists(tempDirectory) && w.ElapsedMilliseconds < 10 * 1000) Thread.Sleep(100);
```
tannergooding: nit: Could you break this up into multiply lines

It might also be good to have this use a back-off factor and a retry count, rather than just sleeping for 100ms intervals.

Having 10 retries, a starting wait time of 10ms, and a back-off factor of 2 will give us ~10s to complete 10 tries.

AbhitejJohn: So the Directory.Delete itself does not have any problems. Its not that it does not delete the directory, its just that it does not necessarily delete it immediately after it comes out of that call. The check for existence should ensure that condition holds true, hence dont see a need for a retry as such. I could add a back-off factor but if there isn't a retry it wouldn't make much of a difference.

AbhitejJohn: @tannergooding : Fixed the spacing.

PR #264, opened 2018-05-29
Run !netcoreapp tests on all platforms

radical:
PR #192 broke running tests on mono (net461) on OSX.
Now:
 - run netcoreapp tests on only Linux
 - run !netcoreapp tests on all platforms

### DIFF
--- a/sdks/RepoToolset/tools/XUnit.targets
+++ b/sdks/RepoToolset/tools/XUnit.targets
@@ -29,7 +29,7 @@
           Inputs="*%(_TestArchitectureItems.Identity)"
           Outputs="*%(_TestArchitectureItems.Identity)"
           Returns="@(_FailedTestRuns)"
-          Condition="$(TargetFramework.StartsWith('netcoreapp')) or '$(OS)' == 'Windows_NT'">
+          Condition="$(TargetFramework.StartsWith('netcoreapp')) or '$(OS)' == 'Windows_NT' or '$(MSBuildRuntimeType)' == 'Mono'">
 
     <PropertyGroup>
       <_TestArchitecture>%(_TestArchitectureItems.Identity)</_TestArchitecture>


radical:
@tmat ^ PR #192 broke running msbuild tests on mono/osx.
cc @rainersigwald

radical:
When building for mono we build with mono's msbuild. And under mono, `Process` (and thus Exec) can directly handle managed assemblies, so you don't need to detect mono. This was working already but it broke after the change in PR #192 and msbuild/upstream moving to a recent version of the tools.

radical:
Maybe we can have a property to force running the tests, besides the existing condition?

`$(TargetFramework.StartsWith('netcoreapp')) or '$(OS)' == 'Windows_NT' or $(RoslynToolsRunTests) == 'true'`

or

```
<PropertyGroup>
   <RoslynToolsRunTests Condition="'$(RoslynToolsRunTests)' == '' and !($(TargetFramework.StartsWith('netcoreapp')) or '$(OS)' == 'Windows_NT')">false</RoslynToolsRunTests>
   <RoslynToolsRunTests Condition="'$(RoslynToolsRunTests)' == ''">true</RoslynToolsRunTests>
</PropertyGroup>
...
<Target Name="InnerTest" Condition="'$(RoslynToolsRunTests)' == 'true'">
```

And for the mono build we can set that to `true` in msbuild.

tmat:
>  And under mono, Process (and thus Exec) can directly handle managed assemblies

Can we detect that we are running on msbuild on Mono? What is the value of `$(MSBuildRuntimeType)` in such case?

radical:
> What is the value of $(MSBuildRuntimeType) in such case?`

`"Mono"`. But using a new property control unconditionally running tests, allows consumers of roslyn-tools, like Mono/msbuild, to control when to run/skip tests.

tmat:
Then let's change to the condition to:
`Condition="$(TargetFramework.StartsWith('netcoreapp')) or '$(OS)' == 'Windows_NT' or '$(MSBuildRuntimeType)' == 'mono'"`

radical:
What about adding a property to control this? If we (mono) need more granular control over running tests then we'll need to edit this file again! It sounds generally useful for downstream consumers to me.

tmat:
> What about adding a property to control this?

There are already other means to control when particular tests run or not. This condition is the default and is meant to avoid running tests that can only be run in Windows without the need to handle this in all tests projects that target netXY.

radical:
You can retain the default and only the projects that do want to change that behavior need set this property. And if the conditions that downstream consumer needs aren't a subset of that default, won't get the opportunity to control which tests to run at all.

```
<PropertyGroup>
   <RoslynToolsRunTests Condition="'$(RoslynToolsRunTests)' == '' and !($(TargetFramework.StartsWith('netcoreapp')) or '$(OS)' == 'Windows_NT')">false</RoslynToolsRunTests>
   <RoslynToolsRunTests Condition="'$(RoslynToolsRunTests)' == ''">true</RoslynToolsRunTests>
</PropertyGroup>
```

Anyway, the runtimeType check will be enough for me right now. So, if you don't agree with the above, then I'll add the `$(MSBuildRuntimeType)==Mono` change.

tmat:
Let's not introduce a property that no one currently needs.

radical:
@tmat updated

tmat:
Thanks for fixing this.

PR #266, opened 2018-06-05
Update XliffTasks

tmeschter:
Update to the latest version of XliffTasks. Notably, this version
improves the MSBuild error messages that occur when a .xlf file can't be
loaded as XML. Previously you wouldn't even be told which .xlf file was
problematic; now you get the file name and the line and column of the
error in the XML.


tmeschter:
test this

tmeschter:
retest this please

PR #279, opened 2018-06-21
Specify IsImplicitlyDefined on PackageReferences

AndyGerlicher:
With this metadata, VS will stop prompting about package updates and the Central Package Versioning SDK will not error requiring you to not specify a version in the PackageReference item directly.

See https://github.com/Microsoft/MSBuildSdks and https://github.com/Microsoft/msbuild/pull/3441

### DIFF
--- a/src/RepoToolset/tools/Compiler.props
+++ b/src/RepoToolset/tools/Compiler.props
@@ -7,10 +7,10 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(MSBuildRuntimeType)' == 'Core'">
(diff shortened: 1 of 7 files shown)


AndyGerlicher:
@jainaashish Does this seem reasonable? Should we should recommend PackageReferences that are included from an SDK add this metadata? Or is this not a great idea?

jainaashish:
@AndyGerlicher yes this seems reasonable to me and SDK already passing this metadata as `true` in restore nomination so I'd imagine they're already setting it on `PackageReference` as well. 

Fyi // @rrelyea @nkolev92

PR #280, opened 2018-06-21
add a `version` property to the inserted component

brettfo:
This is to make other internal tooling better able to handle merge conflicts by adding a version for each inserted component.

The `.vsman` files used for insertion will be scanned for a `info/buildVersion` property which will be forwarded into a `version` property in the config JSON file.  See Microsoft/visualfsharp#5226.

### DIFF
--- a/src/RepoToolset/tools/VisualStudio.InsertionManifests.targets
+++ b/src/RepoToolset/tools/VisualStudio.InsertionManifests.targets
@@ -13,9 +13,10 @@
     This target is invoked in a separate phase after all input VSIX files are signed.
     This is important since the manifest contain hashes of the VSIX files.
   -->
-  <Target Name="GenerateVisualStudioInsertionManifests" 
+  <Target Name="GenerateVisualStudioInsertionManifests"
           AfterTargets="Pack"
-          Outputs="%(_StubDirs.Identity)" 
+          Outputs="%(_StubDirs.Identity)"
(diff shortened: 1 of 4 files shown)


src/RepoToolset/tools/VisualStudio.InsertionManifests.targets
```diff
@@ -27,6 +27,7 @@
       <_Args Include="SetupOutputPath=$(VisualStudioSetupInsertionPath)"/>
       <_Args Include="ComponentIntermediateOutputPath=$(VisualStudioSetupIntermediateOutputPath)$(_ComponentName)\"/>
       <_Args Include="SwixBuildPath=$(NuGetPackageRoot)microbuild.plugins.swixbuild\$(MicroBuildPluginsSwixBuildVersion)\"/>
+      <_Args Include="ManifestBuildVersion=$(ManifestBuildVersion)" />
```
brettfo: Consumers of the repo toolset will need to specify this value (see dotnet/project-system#3669) and update their toolset reference.  The change is such that if only one of these changes (the property or the toolset update) is accepted, then nothing will break, but the `.vsman` files just won't have the new `info/buildVersion` property.

tmat: Why is this not set directly to `VsixVersion`?

PR #290, opened 2018-07-11
Allow 20 minutes for the VSTS DownloadArtifactsAsync instead of 10

dpoeschl:
Changes that just increase timeouts aren't great, but it sounds like this is the right option for now. This step often takes a bit longer than 10 minutes, and I have to sit there respinning over and over just hoping.


dpoeschl:
I also talked to @brettfo about this a few days ago and we thought this would be reasonable.

PR #293, opened 2018-07-17
add web job to auto-merge some auto-created PRs

brettfo:
This is my attempt to re-create (and add to source control) an Azure function that I accidentally deleted when I was trying to add it to source control to make sure it never got accidentally deleted.  See #292 for a similar scenario.

@agocke Can you take a look at `run.csx` and see if that looks similar to what existed before?

I validated this by creating a new Azure Function host `roslyn-infra-functions-playground` and verified that it ran.


src/AzureFunctionPackage/Functions/GitHubAutoMergePRs/GitHubAutoMergePRs.csproj
```diff
@@ -0,0 +1,34 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net46</TargetFramework>
+  </PropertyGroup>
+
+  <!--
+
+  The dependencies specified in `project.json` are the real ones.  These simply exist so that the generated dependency
+  graph (`project.assets.json`) can be copied as a build step.
```
jasonmalinowski: Do we have a plan to remove this duplication? Is this something we also invented or is the standard SDK?

brettfo: Azure Functions currently only support `project.json` scenarios, but since that's not something we can easily build and test locally due to the 2.x SDK completely removing `project.json` support, this was the next best thing.

PR #302, opened 2018-07-25
Replace special casing of feature branches

agocke:
Feature branches currently have a special rule that treats any
test failure as a blocker. This is because many feature branches
don't have required tests, so otherwise all test failures would be
ignored. Rather than special casing feature branches, it's better
to recognize when there are no required branches and treat that
as the signifier instead


src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -283,22 +283,22 @@ git commit
             var testStatusBody = JObject.Parse(await testStatusResponse.Content.ReadAsStringAsync());
             var statusDict = testStatusBody["statuses"].Select(t => ((string)t["context"], "success" == (string)t["state"]))
                 .ToDictionary(t => t.Item1, t => t.Item2);
-            foreach (var test in requiredTests)
+
+            // If there are no required tests, treat *any* test failure as a blocker
```
jasonmalinowski: This is a good fail-safe, I like it.

PR #303, opened 2018-07-26
Remove post sign validation

jaredpar:
The post sign validation phase had very questionable value in the
SignTool. The intent was to verify assemblies were actually signed by
the sign process by calling into the CLR to verify their strong name
signature. That is flawed for a couple of reasons:

1. Strong naming is only part of the verification process hence the
verification is incomplete.
1. The exit code of signing actually tells us if signing succeeded or
not. Hence doing this validation is essentially not trusting the tooling
return code which is probably over paranoid.

Additionally this was causing problems when signing native binaries. To
the sign tool it's impossible to tell whether a.exe is a managed or
native binary. Strong name validation will always fail for a.exe even
when it is properly cert signed. Hence simplyfiying things here by
removing the check altogether.

### DIFF
--- a/src/SignTool/SignTool/BatchSignUtil.cs
+++ b/src/SignTool/SignTool/BatchSignUtil.cs
@@ -57,10 +57,7 @@ namespace SignTool
             RemovePublicSign(textWriter);
 
             // Next sign all of the files
-            SignFiles(contentMap, zipDataMap, textWriter);
-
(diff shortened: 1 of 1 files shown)


jaredpar:
Note: we should do this in arcade as well. At the moment though it's blocking users who are still using the EXE version of the tool. Hence have to make the fix here.

PR #305, opened 2018-08-02
Abandon insertion if there were only whitespace changes

agocke:


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Git.cs
```diff
@@ -238,15 +239,67 @@ private static void StageFiles(BuildVersion newRoslynVersion,CancellationToken c
...
+                        while ((line = reader.ReadLine()) != null)
+                        {
+                            if (line.StartsWith("---") || line.StartsWith("+++"))
+                            {
+                                continue;
+                            }
+
+                            if (line.StartsWith("+"))
+                            {
+                                adds.Append(line.TrimStart('+').Trim());
```
jasonmalinowski: .TrimStart won't trim just the first + but rather all of them, so if you had a line going from + to - (as actual text content), this will mis-fire. :smile:

agocke: Good point. I will just trim the first character.

PR #310, opened 2018-08-08
Add a `-diag` switch to sign tool to show more verbose output

natemcmaster:
Make it easier to gather more detailed information when something is going wrong.

### DIFF
--- a/src/SignTool/README.md
+++ b/src/SignTool/README.md
@@ -2,15 +2,15 @@
 
 This is a batch signing and verification tool for Microbuild environments.  The tool is run as a post-build step and driven by a declarative configuration file.  The high level features of the tool are:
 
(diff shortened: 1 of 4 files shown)


natemcmaster:
My editor did what it was told to do.
https://github.com/dotnet/roslyn-tools/blob/3153452c268f47a8282309d6ecb759b0c7697d2c/.editorconfig#L18-L24

jmarolf:
@natemcmaster binary logging should already work for signtool, that's about as diagnostic as you can get. What is the binlog capturing?

natemcmaster:
I tried that, but the problem was that that msbuild invocation never terminated, so the binary log was corrupt and couldn't be read. I had to re-create the generated "build.proj" file and invoke MSBuild manually with `/v:diag` to figure out what was happening.

natemcmaster:
Thanks for the review @jmarolf. I don't have write permissions, so someone else will have to merge this for me.

PR #312, opened 2018-08-16
Honor value for UsingToolMicrosoftNetCompilers

AndyGerlicher:
Only set to true when the property is not set.


JoeRobich:
The changes in this PR have been superseded by dotnet/arcade - https://github.com/dotnet/arcade

PR #315, opened 2018-08-20
Resolve mscorlib resource assembly

cston:
Resolve mscorlib satellite assembly before adding an `AssemblyResolve` handler to avoid a cycle resolving the assembly if an exception is thrown in the handler.


cston:
@jmarolf, @jaredpar please review.

PR #316, opened 2018-08-21
allow insertion tool to only update assemblt versions

jmarolf:
If an insertion has a DevDivInsertionFiles\DependentAssemblyVersions.csv file allow it to be used to update binding redirects

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
@@ -182,7 +182,10 @@ namespace Roslyn.Insertion
                     {
                         return;
                     }
+                }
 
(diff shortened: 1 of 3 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -182,7 +182,10 @@ public static partial class RoslynInsertionTool
                     {
                         return;
                     }
```
brettfo: Should there be a `retainBuild = true;` here?

jmarolf: nope there is a `!` in front.  If we get here then file copy failed and we should bail out

PR #326, opened 2018-09-13
Update RIT to kick off the correct RPS runs (both old & new) when insertions are created

dpoeschl:
The regular queue's name got marked as `[deprecated]` which broke things. Trying to add the new queue while we're here.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
@@ -339,7 +339,7 @@ namespace Roslyn.Insertion
 
                     try
                     {
-                        await QueueBuildPolicy(pullRequest, "VAL build with RPS");
+                        await QueueBuildPolicy(pullRequest, "[Deprecated] ValBuild RPS");
                     }
                     catch (Exception ex)
                     {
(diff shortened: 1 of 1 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -339,13 +339,22 @@ public static partial class RoslynInsertionTool
...
                     catch (Exception ex)
                     {
-                        Log.Error($"Unable to create a validation build for '{pullRequest.SourceRefName}'");
+                        Log.Error($"Unable to create a deprecated validation build for '{pullRequest.SourceRefName}'");
+                        Log.Error(ex);
+                    }
+
+                    try
+                    {
+                        await QueueBuildPolicy(pullRequest, "CloudBuild - RPS");
```
dpoeschl: Not sure what's right here.  There's "CloudBuild - Request RPS" and also "CloudBuild RPS Status" and I'm not sure if the "Status" suffix is included or which one is right?

brettfo: I don't think 'Status' is part of the title because it's listed on PRs as 'CloudBuild - Request RPS succeeded' (or failed).

dpoeschl: @brettfo So then it should be "CloudBuild - Request RPS" (instead of what I have without the "Request")?

image

PR #330, opened 2018-09-19
Update insertion graph now that 15.9 P3 has snapped

dpoeschl:


dpoeschl:
FYI @Shyam-Gupta @brettfo @jmarolf

dpoeschl:
Already copied this to the server and am running it.

PR #355, opened 2018-10-01
Generate the build.proj file in the intermediate output path

natemcmaster:
When signtool.exe is invoked in parallel on the same machine, there is contention for access to $AppBaseDir/build.proj. This moves the build.proj file into the intermediate output path instead to avoid this contention.

cc @jmarolf


natemcmaster:
@tmat can you take a look?

PR #356, opened 2018-10-01
Improve Insertion Titles

dpoeschl:
Include the originating branch name & build number

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -39,25 +39,25 @@ namespace Roslyn.Insertion
 
         private static TfsTeamProjectCollection ProjectCollection => LazyProjectCollection.Value;
 
-        private static GitPullRequest CreatePullRequest(string sourceBranch, string targetBranch, string description)
+        private static GitPullRequest CreatePullRequest(string sourceBranch, string targetBranch, string description, string buildToInsert)
         {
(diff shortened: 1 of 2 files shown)


dpoeschl:
So, I haven't actually tested this. Need to figure out how to build & run on my machine. Anyway, tagging @brettfo for an early review.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -44,7 +44,7 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
             Log.Trace($"Creating pull request sourceBranch:{sourceBranch} targetBranch:{targetBranch} description:{description}");
             return new GitPullRequest
             {
-                Title = $"{Options.InsertionName} Insertion Into {Options.VisualStudioBranchName} ",
+                Title = $"{Options.InsertionName} '{Options.BranchName}/{Options.SpecificBuild}' Insertion into {Options.VisualStudioBranchName} ",
```
brettfo: You'll want to pass through and consume `buildToInsert` from [here](https://github.com/dotnet/roslyn-tools/blob/e22a64f07f5c84b261ce57c10ef29f2757d97a79/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs#L110) (and also line 100 above) (or `buildVersion`).  `Options.SpecificBuild` is a command-line override that will only be used in manual cases.

dpoeschl:
FYI @jamshedd that this is being worked on.

dpoeschl:
@brettfo I've updated it with your suggestions. Can you give it another look?

PR #365, opened 2018-10-08
RIT: Skip builds tagged with 'DoesNotRequireInsertion'

dpoeschl:


dpoeschl:
Do we need to have this be per-target-branch instead of just flagging the build across the board.

dpoeschl:
Tagging @jasonmalinowski since we talked about this earlier today.

jasonmalinowski:
@dpoeschl It is technically a per-target-branch concept, yes. I could imagine us wanting a per-branch flag because we might say something like "we don't need to insert this build before the division snaps branches, but once it does we want to insert it somewhere else."

jasonmalinowski:
Given the ease of including the branch name into the tag....I'd just go with that.

dpoeschl:
Okay, will do.

dpoeschl:
Okay, it's now per-branch.

dpoeschl:
Gonna give it a spin.

PR #373, opened 2018-10-16
Allow insertion tool to run from an Azure DevOps release.

brettfo:
This removes the requirement of our internal machine which only exists (as far as I can tell) to perform regular VS insertions.  With these changes I've confirmed that all regular insertions can be performed with an Azure DevOps release definition ([here](https://dev.azure.com/devdiv/DevDiv/_releases2?definitionId=1001), specifically `Release-22`).  Insertions created with these changes can be found [here](https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequests?creatorId=9f64bc2f-479b-429f-a665-fec80e130b1f&_a=active) by looking for PRs that start with "[TEST ONLY]".

Going commit-by-commit might be easiest, but essentially all logging is now redirected to the console (since ADO captures all console output) and there is some plumbing to pass through secrets.


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -299,102 +292,84 @@ public static partial class RoslynInsertionTool
...
                     {
-                        Log.Error("Unable to create a validation build: no pull request.");
-                        return;
+                        Console.WriteLine("Unable to create a validation build: no pull request.");
+                        return false;
                     }
 
                     try
                     {
-                        await QueueBuildPolicy(pullRequest, "CloudBuild - Request RPS");
```
dpoeschl: This, in particular, will need to be reverted. We should also look through the rest of this to be sure other changes weren't lost.

PR #378, opened 2018-10-19
Don't trigger deprecated RPS run

shyamnamboodiripad:


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -323,16 +323,6 @@ public static partial class RoslynInsertionTool
...
-                    }
-                    catch (Exception ex)
-                    {
-                        Console.WriteLine($"Unable to create a deprecated validation build for '{pullRequest.SourceRefName}'");
-                        Console.WriteLine(ex);
-                    }
-
                     try
                     {
                         await QueueBuildPolicy(pullRequest, "CloudBuild - RPS");
```
shyamnamboodiripad: @dpoeschl @brettfo I don't see an entry for "CloudBuild - DDRITs". Do you know Is that triggered implicitly  somehow or is that being triggered via the deprecated policy that I am removing above?

dpoeschl:
```suggestion
                        await QueueBuildPolicy(pullRequest, "CloudBuild - Request RPS");
```

dpoeschl:
```suggestion
                         await QueueBuildPolicy(pullRequest, "CloudBuild - Request RPS");
```

shyamnamboodiripad: Thanks! Have fixed that.

dpoeschl:
The need for that whitespace fixup is a bug in the suggested changes feature. 😞

Issue #396, opened 2018-10-26
[RIT] Allow an existing PR to be appended to instead of replaced.


brettfo:
Currently the `OverwriteExistingPr.ps1` script (and matching Release Definition) always do a force push to the branch specified by the PR.

An additional `AppendExistingPr.ps1` script and the matching Release Definition should be added for the times when it's requested that an existing PR has important work but still needs an insertion.

AbhitejJohn:
This has been addressed as part of #426.

PR #426, opened 2018-12-01
Update Existing PR

AbhitejJohn:
Added in an ability to update an existing PR as a fix for #396. If one chooses to overwrite it would now be a flag one can specify with the update schematics.

Also fixed the possibility of resetting to a stale base branch with a Fetch.

This might need a change to certain Release definitions - if someone can point me to these I'll go fix them up as well.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -281,14 +281,13 @@ namespace Roslyn.Insertion
             if (Directory.Exists(tempDirectory))
             {
                 // Be judicious and clean up old artifacts so we do not eat up memory on the scheduler machine.
-                //Directory.Delete(tempDirectory, recursive: true);
+                Directory.Delete(tempDirectory, recursive: true);
 
-                //// Sometimes a creation of a directory races with deletion since at least in .net 4.6 deletion is not a blocking call.
-                //// Hence explictly waiting for the directory to be deleted before moving on.
-                //Stopwatch w = Stopwatch.StartNew();
+                // Sometimes a creation of a directory races with deletion since at least in .net 4.6 deletion is not a blocking call.
+                // Hence explictly waiting for the directory to be deleted before moving on.
+                Stopwatch w = Stopwatch.StartNew();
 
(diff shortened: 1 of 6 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -281,13 +281,14 @@ private static async Task<string> DownloadArtifactsAsync(BuildHttpClient buildCl
...
                 // Be judicious and clean up old artifacts so we do not eat up memory on the scheduler machine.
-                Directory.Delete(tempDirectory, recursive: true);
-                
-                // Sometimes a creation of a directory races with deletion since at least in .net 4.6 deletion is not a blocking call.
-                // Hence explictly waiting for the directory to be deleted before moving on.
-                Stopwatch w = Stopwatch.StartNew();
+                //Directory.Delete(tempDirectory, recursive: true);
 
-                while (Directory.Exists(tempDirectory) && w.ElapsedMilliseconds < 20 * 1000) Thread.Sleep(100);
+                //// Sometimes a creation of a directory races with deletion since at least in .net 4.6 deletion is not a blocking call.
```
brettfo: Comment cleanup.

src/RoslynInsertionTool/scripts/UpdateExistingPr.ps1
```diff
@@ -16,7 +16,8 @@ param([string] $enlistmentPath,
       [string] $updateAssemblyVersions,
       [string] $updateCoreXTLibraries,
       [string] $visualStudioBranchName,
-      [string] $writePullRequest)
```
brettfo: Just a note to remember to update the Azure DevOps release definition when this is in to take into account the script name change.  I can help with this if you need.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
AbhitejJohn: Sorry, this was not meant to be. I'll undo this change.

src/RoslynInsertionTool/scripts/UpdateExistingPr.ps1
AbhitejJohn: yes, that would be helpful.

AbhitejJohn: Synced up offline. Will make a change.

PR #429, opened 2018-12-03
Allow for branch under jobs node

jaredpar:
This change allows for the branch specifier to live under the jobs node
in addition to the root of the YAML file.


jaredpar:
@jmarolf how do we get a new NuPkg build of this package so we can get it into dotnet/roslyn?

PR #431, opened 2018-12-05
Updating the overwritePR flag to be a switch instead.

AbhitejJohn:


AbhitejJohn:
This is a follow up to #426 . I'll go ahead and merge this in to get releases unblocked - this should be low risk. I'd be happy to address any concerns after the merge.

PR #432, opened 2018-12-07
Support for Arcade insertion artifacts

tmat:
Abstracts artifact layout and implements it for pre-Arcade drop structure and Arcade structure.


tmat:
@jasonmalinowski @jmarolf @brettfo @dotnet/roslyn-infrastructure PTAL

PR #439, opened 2018-12-11
Fix logic determining valid build artifacts

tmat:


tmat:
@genlu @JoeRobich PTAL

PR #441, opened 2018-12-18
Fix insertions for LUT and VSUT

shyamnamboodiripad:
For LUT and VSUT:
- Remove stale insertions
- Add Preview 2 insertions


src/RoslynInsertionTool/scripts/VS-Insertion.ps1
```diff
@@ -49,18 +49,14 @@ Do-Insertion -component "F#" -queueName "FSharp-Signed" -fromBranch "dev16.0" -t
 
 # Dev15.9 Preview 4
 Do-Insertion -component "Roslyn"            -queueName "Roslyn-Signed"         -fromBranch "dev15.9.x-vs-deps" -toBranch "rel/d15.9" -insertToolset "false" -insertDevdiv "false" -queueValidation "true" -updatecorextlibraries "true"
-Do-Insertion -component "Live Unit Testing" -queueName "TestImpact-Signed"     -fromBranch "dev15.9-preview3"  -toBranch "rel/d15.9" -insertCore "false"    -insertDevdiv "false" -queueValidation "true"
```
333fred: Why remove these? We'll have to redo them if any services fixes are needed.

shyamnamboodiripad: I think we would just kick off a one-off insertion in that case. The branches for servicing (both repo-side and VS-side) are going to be different from what is above - so we would have to edit this regardless...

PR #442, opened 2019-01-04
Do not try to rollback changes on an empty commit

JoeRobich:
We've had numerous incomplete runs of the insertion tool because of a NRE during rollback following an EmptyCommitException. It does not seem necessary to rollback in the case of an empty commit since there should be no changes.


JoeRobich:
@brettfo please take a look

PR #450, opened 2019-01-16
Add an insertion for dev16.1-preview1-vs-deps to lab/ml

JoeRobich:


JoeRobich:
The lab/ml branch has been updated and should be ready. /cc @jaredpar @jasonmalinowski

PR #464, opened 2019-02-28
Use ADO APIs to create placeholder PRs

agocke:


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Git.cs
```diff
@@ -190,21 +193,51 @@ private static Branch SwitchToBranchAndUpdate(string branchToSwitchTo, string ba
...
+                Commits = new[] {
+                    new GitCommitRef()
+                    {
+                        Comment = $"PLACEHOLDER INSERTION FOR {Options.InsertionName}",
+                        Changes = new GitChange[]
+                        {
+                            new GitChange()
+                            {
+                                ChangeType = VersionControlChangeType.Delete,
+                                Item = new GitItem() { Path = "/Init.ps1" }
```
chsienki: what happens if we ever introduce a file called /Init.ps1?

Is git ok with creating a commit that deletes a non existent file?

agocke: We have a file called `Init.ps1` -- it's the main file! I include this change so that the tests fail and the PR doesn't accidentally get merged. If we ever end up changing the name of the entry script, this might fail, but that seems like a rare/unlikely change.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Git.cs
```diff
@@ -190,21 +193,51 @@ private static Branch SwitchToBranchAndUpdate(string branchToSwitchTo, string ba
...
+
+            var refs = await gitClient.GetRefsAsync(
+                repository.Id,
+                filter: $"heads/{Options.VisualStudioBranchName}",
+                cancellationToken: cancellationToken);
+            GitRef sourceBranch = refs.Single();
+
+            var branchName = GetNewBranchName();
+
+            _ = await gitClient.CreatePushAsync(new GitPush()
```
RikkiGibson: suppressing an unused value warning?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Git.cs
```diff
@@ -190,21 +193,51 @@ private static Branch SwitchToBranchAndUpdate(string branchToSwitchTo, string ba
...
+                        {
+                            new GitChange()
+                            {
+                                ChangeType = VersionControlChangeType.Delete,
+                                Item = new GitItem() { Path = "/Init.ps1" }
+                            },
+                        }
+                    },
+                },
+            }, repository.Id, cancellationToken: cancellationToken);
```
RikkiGibson: nit: consider removing the argument label

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Git.cs
agocke: yup

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Git.cs
agocke: Can't: there's an intermediate optional argument that I don't want to provide a value for.

PR #468, opened 2019-03-07
Disable localization for roslyn-tools

agocke:
Localization produces a lot of large resource assemblies that we don't need, and increases the time needed to download artifacts in deployment.


tmat:
LGTM. It seems though that we should also set `IsShipping` false in https://github.com/dotnet/roslyn-tools/blob/master/src/Directory.Build.props.

agocke:
@tmat Isn't that set already? https://github.com/dotnet/roslyn-tools/blob/master/src/Directory.Build.props#L11

tmat:
I missed that. Then it's odd that localization would kick in. It should only apply to shipping projects.

PR #474, opened 2019-03-14
Support creating weekly github auto-merges

JoeRobich:
Frequent requests to fix merge conflicts or build warnings in merges to feature branches can be disruptive. Allow feature branches to opt-in to a weekly merge.

### DIFF
--- a/src/AzureFunctionPackage/Functions/GitHubCreateMergePRs/run.csx
+++ b/src/AzureFunctionPackage/Functions/GitHubCreateMergePRs/run.csx
@@ -40,7 +40,7 @@ private static async Task MakeGithubPr(
     }
 }
 
-private static async Task RunAsync(ExecutionContext context, bool isManualRun)
+private static async Task RunAsync(ExecutionContext context, bool isAutomatedRun)
 {
     var gh = new GithubMergeTool.GithubMergeTool("dotnet-bot@users.noreply.github.com", await GetSecret("dotnet-bot-github-auth-token"));
     var configPath = Path.Combine(context.FunctionDirectory, "config.xml");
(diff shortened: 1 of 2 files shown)


src/AzureFunctionPackage/Functions/GitHubCreateMergePRs/run.csx
```diff
@@ -53,6 +53,13 @@ private static async Task RunAsync(ExecutionContext context)
         {
             var fromBranch = merge.Attribute("from").Value;
             var toBranch = merge.Attribute("to").Value;
+
+            var frequency = merge.Attribute("frequency")?.Value ?? "daily";
```
agocke: I think the default merge frequency should be higher. It would be nice if we could get main branches done probably twice a day to prevent big merge conflicts.

JoeRobich: Yeah. I think my choice of work there is bad. Nothing would stop this from running multiple times a day, or even "weekly" merges multiple times on a Sunday

JoeRobich: @brettfo Do you have any ideas on how I could do this better?

dpoeschl: I think I still want to be able to go run the function manually at any time and have it generate PRs for me, even if it's on a Wednesday.

JoeRobich: Been thinking about the TimerInfo passed into the Run method. Perhaps we can use it to determine if we are being manually run.

JoeRobich: After playing with Azure Functions a bit I worked out how to determine if it is a manual run. I also removed the poorly named "daily" default.

@agocke Looking at the function.json for the CreateMergePRs function. It is running every 12 hours https://github.com/dotnet/roslyn-tools/blob/master/src/AzureFunctionPackage/Functions/GitHubCreateMergePRs/function.json

PR #480, opened 2019-04-03
Skip dotnet tools

wli3:
[Packaged shim](https://github.com/dotnet/cli/blob/master/Documentation/general/signing-global-tool-packages.md) has PackageVersion
embedded. Repack a dotnet tool will create an invalid package.


wli3:
@tmat for review

src/NuGetRepack/Common/NuGetVersionUpdater.cs
```diff
@@ -215,6 +228,13 @@ private static void LoadPackages(IEnumerable<string> packagePaths, Dictionary<st
                         }
                     }
 
+                    if (isDotnetTool)
```
wli3: It will has some waste compute but the alternative of break 2 layer of look is goto...

src/NuGetRepack/Common/NuGetVersionUpdater.cs
```diff
@@ -163,6 +164,16 @@ private static void LoadPackages(IEnumerable<string> packagePaths, Dictionary<st
...
 
+                            if (packageVersionStr == null)
+                            {
+                                throw new InvalidDataException($"'{packagePath}' has invalid nuspec: missing 'version' element");
+                            }
+
+                            if (packageVersionStr == null)
+                            {
+                                throw new InvalidDataException($"'{packagePath}' has invalid nuspec: missing 'version' element");
+                            }
```
tmat: These seem to be duplicated?

src/NuGetRepack/Common/NuGetVersionUpdater.cs
```diff
@@ -250,6 +270,25 @@ private static void LoadPackages(IEnumerable<string> packagePaths, Dictionary<st
...
+        private static bool IsDotnetTool(string nuspecXmlns, XElement metadata)
+        {
+            var packageTypesElement = metadata.Element(XName.Get("packageTypes", nuspecXmlns));
+            if (packageTypesElement != null)
+            {
+                foreach (var packageType in packageTypesElement.Elements(XName.Get("packageType", nuspecXmlns)) ?? Array.Empty<XElement>())
+                {
+                    var name = packageType.Attribute("name");
+
+                    if (name != null && name.Value != null && string.Equals(name.Value, "DotnetTool", StringComparison.OrdinalIgnoreCase))
```
tmat: >name != null && name.Value != null && string.Equals(name.Value [](start = 24, length = 62)

Nit: Can be simplified to `string.Equals(name?.Value, "DotnetTool", StringComparison.OrdinalIgnoreCase)`.

tmat:
Looks good module couple of details.

src/NuGetRepack/Common/NuGetVersionUpdater.cs
wli3: oops, removed

src/NuGetRepack/Common/NuGetVersionUpdater.cs
wli3: Used suggested

wli3:
I don't have the access to merge, please help me merge it when you think it is good

PR #481, opened 2019-04-04
Add an extra message to generated PRs created from manual runs

RikkiGibson:


RikkiGibson:
Will probably merge after the snap out of an abundance of caution.

PR #491, opened 2019-04-24
Remove dev16.1-preview2 from Roslyn github codeflow

dpoeschl:


dpoeschl:
FYI @jasonmalinowski

src/AzureFunctionPackage/Functions/Common/config.xml
```diff
@@ -3,22 +3,19 @@
     <!-- Roslyn branches (non-vs-deps) (flowing between releases) -->
     <merge from="dev15.0.x" to="dev15.9.x" />
     <merge from="dev15.9.x" to="dev16.0" />
-    <merge from="dev16.0" to="dev16.1-preview2" />
-    <merge from="dev16.1-preview2" to="dev16.1" />
```
jasonmalinowski: If we take any Preview 2 hotfixes we'll just do it manually?

dpoeschl: I think that will be fine, yes.

PR #501, opened 2019-05-06
Send both base and head shas when merging

JoeRobich:
GitHub V3 API expects both head and base shas https://developer.github.com/v3/repos/merging/#input


JoeRobich:
Seems I was looking at the wrong documentation. The current code is correct https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button

PR #507, opened 2019-05-10
Add testimpact repo branches

shyamnamboodiripad:


src/AzureFunctionPackage/Functions/Common/VstsPrConfig.xml
```diff
@@ -1,8 +1,7 @@
 <config>
   <repo name="VSUnitTesting">
-    <merge from="dev15.9.x" to="master" />
-    <merge from="dev16.0-preview1" to="master" />
-    <merge from="dev16.0-preview2" to="master" />
-    <merge from="dev16.0.x" to="master" />
+    <merge from="dev15.9.x" to="dev16.0.x" />
```
AbhitejJohn: The to on all of these should ideally be master so its a conscious choice when we decide to ship service updates to already released branches.

shyamnamboodiripad: I just copy pasted the pattern that all other repos were following... Why isn't that the case for Roslyn, F# etc.?

PR #515, opened 2019-06-03
Roslyn - No need to create insertion PRs for rel/d16.2

JoeRobich:


JoeRobich:
> this is because all insertions will be via manual dummy PRs that get overwritten right?

@RikkiGibson That's right.

PR #530, opened 2019-07-15
Handle exceptions on attempted merges

ryzngard:
Add try/catch around attempted merges and log failures.

### DIFF
--- a/src/AzureFunctionPackage/Functions/GitHubAutoMergePRs/run.csx
+++ b/src/AzureFunctionPackage/Functions/GitHubAutoMergePRs/run.csx
@@ -33,22 +33,32 @@ private static async Task RunAsync(ExecutionContext context)
         {
             var prIdentifier = $"{owner}/{name}:{pr}";
             Log.Info("Checking " + prIdentifier);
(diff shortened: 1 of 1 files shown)


src/AzureFunctionPackage/Functions/GitHubAutoMergePRs/run.csx
```diff
@@ -33,22 +33,32 @@ private static async Task RunAsync(ExecutionContext context)
...
-            else if (mergeError != null)
-            {
-                Log.Error($"Unable to auto-merge PR '{prIdentifier}': {await mergeError.Content.ReadAsStringAsync()}");
-            }
-            else
-            {
-                Log.Error($"Unable to auto-merg PR '{prIdentifier}' for unknown reason.");
+                // If a specific merge fails, we don't want to fail all merges. Log the exception
+                // and continue trying to merge other PRs 
+                Log.Error(ex);
```
ryzngard: I wasn't able to run this locally, is this the right way to log exceptions or is there something more detailed?

jasonmalinowski: 🤷‍♂

PR #533, opened 2019-07-17
Log.Error doesn't have an overload that takes an exception

ryzngard:
Use `ex?.ToString()` instead to output the error.


ryzngard:
:(

PR #538, opened 2019-08-06
Removed NuGet.Repack now that it lives in Arcade

JoeRobich:


JoeRobich:
@tmat Please take a look

PR #539, opened 2019-08-09
Create draft insertion PRs with Roslyn Insertion Tool

JoeRobich:
Specify "/createdraftpr=true" to create your insertion as a draft. Additionally specify, "/queuevalidationbuild=true" to start all the build policies.

Created this insertion PR with RIT to verify https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/232168


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -44,7 +44,8 @@ private static GitPullRequest CreatePullRequest(string sourceBranch, string targ
                 Title = $"{prefix}{Options.InsertionName} '{Options.BranchName}/{buildToInsert}' Insertion into {Options.VisualStudioBranchName}",
                 Description = description,
                 SourceRefName = sourceBranch,
-                TargetRefName = targetBranch
+                TargetRefName = targetBranch,
+                IsDraft = true,
```
RikkiGibson: What's the effect in practice of making the insertion PR a draft? Reduces notification noise?

JoeRobich: It does. Need to remember who was asking for this. Maybe @jasonmalinowski remembers

JoeRobich: A thing to think about is that when an insertion pr is marked as a draft, none of the policies run. Manually queuing a policy still won't run it.

JoeRobich: So it seemed like this would be useful for DummyPRs, but you also can't approve a draft PR which makes it less than ideal

jasonmalinowski: I think we did want draft support so anybody making insertions for their own uses (to run RPS) could be marked as draft so nobody gets excited pressing merge buttons. "Normal" insertions would not use draft.

JoeRobich: I could not get any policies to run while in Draft state. Cloudbuild, DDRIT, Symbol check, etc....

JoeRobich: I came back around to this and draft PRs can now launch policies to run. They do not start automatically so I will look into starting them with RIT.

JoeRobich: Updated so that specifying /QueueValidationBuild=true will start all the checks for draft PRs now.

JoeRobich:
@RikkiGibson Please take another look

RikkiGibson:
a few things:

1. have you verified the expected behavior for UpdateExistingPr and OverwriteExistingPr? I assume that Update would leave the draft status as-is and Overwrite would overwrite the draft status..?
2. which of our scripts, if any, should pass this flag? should we just remember to pass the flag when queueing a OneOffInsertion for a dev branch?

JoeRobich:
> a few things:
> 
> 1. have you verified the expected behavior for UpdateExistingPr and OverwriteExistingPr? I assume that Update would leave the draft status as-is and Overwrite would overwrite the draft status..?

Updated the Options.Valid check. Will now report invalid and give a message if CreateDraftPR is true and UpdatingExistingPR. Verfied Update does not affect the DraftPR state. Verified that Overwrite can revert a published PR back to a Draft PR.

> 2. which of our scripts, if any, should pass this flag? should we just remember to pass the flag when queueing a OneOffInsertion for a dev branch?

OneOffInsertion and Overwrite should have the option. You cannot approve a PR while it is unpublished so it doesn't work for the DummyPR use case.

PR #549, opened 2019-08-23
Invert the merge flow for dev16.4-preview1

agocke:
master is taking over, so the merges should go the other way until we can delete the branch


src/AzureFunctionPackage/Functions/Common/config.xml
```diff
@@ -17,7 +17,7 @@
     <merge from="release/dev16.1-vs-deps" to="release/dev16.2" />
     <merge from="release/dev16.3-preview2-vs-deps" to="release/dev16.3-vs-deps" />
     <merge from="release/dev16.3-vs-deps" to="master-vs-deps" />
-    <merge from="master-vs-deps" to="release/dev16.4-preview1-vs-deps" />
+    <merge from="release/dev16.4-preview1-vs-depss" to="master-vs-deps" />
```
RikkiGibson: typo?

agocke: Yup, thanks

PR #556, opened 2019-10-01
Updating the new RPS policy name

AbhitejJohn:
RPS isn't triggering on the insertion builds because of a policy name change that was made [here](https://devdiv.visualstudio.com/Engineering/_git/PullRequestBot/pullrequest/204633?_a=files). Updating our scripts to reflect the new name.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
@@ -336,7 +336,7 @@ namespace Roslyn.Insertion
 
                     try
                     {
-                        await QueueBuildPolicy(pullRequest, "CloudBuild - Request RPS");
+                        await QueueBuildPolicy(pullRequest, "Request Perf DDRITs");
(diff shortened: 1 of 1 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -336,7 +336,7 @@ public static partial class RoslynInsertionTool
 
                     try
                     {
-                        await QueueBuildPolicy(pullRequest, "CloudBuild - Request RPS");
```
shyamnamboodiripad: @AbhitejJohn I also see `Cloud Build - RPS`. But the comment in the PR says we should be running `Request Perf DDRITs`. Not sure if the `Cloud Build - RPS` would give us faster turnaround?

AbhitejJohn: You can't actually trigger `CloudBuild-RPS`. It looks more like a status indicator to me.

PR #561, opened 2019-10-15
Refactor github merge creation tool and add pipeline to run on schedule.

dibarbet:
1.  Move the github merge creation azure functions into a .netcore project that we can run easily.
     a.  The only change is adding the dryrun flag + cmd line arguments to the script, rest is just copied.
2.  Added dryrun flag which will run the tool but not create any pull requests.
3.  Added pipeline to run the merge tool on a schedule

Did a run with only a couple branches in the config
PR created - https://github.com/dotnet/roslyn/pull/39282
Build - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3142839


dibarbet:
@RikkiGibson I've done my best to separate into multiple commits, let me know if it helps.

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -1,12 +1,92 @@
-﻿using System;
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
-namespace GitHubCreateMergePRs
+#r "System.Xml.Linq"
```
RikkiGibson: I see, so here one of our .csx files is being adapted as the main file for a console app.

shyamnamboodiripad:
@Cosifne @AbhitejJohn probably no changes are needed at the moment but this may impact https://github.com/dotnet/roslyn-tools/tree/master/src/AzureFunctionPackage/Functions/VstsCreateMergePRs. We are not using that at the moment - but I think we'd want to align that with the changes in the current PR at some point...

src/GitHubCreateMergePRs/Program.cs
dibarbet: Yep basically just created a project from the csx files so I could invoke it with dotnet run.

merge-pr-pipeline.yml
```diff
@@ -0,0 +1,31 @@
...
+variables:
+# Variable group for the bot secret.
+- group: RoslynMergeBot
+- name: isDryRun
+  value: false
+- name: isAutomated
+  value: true
+
+schedules:
+- cron: "0 */12 * * *"
```
RikkiGibson: what does this mean? 2x daily?

dibarbet: Yep, it's the same cron as in https://github.com/dotnet/roslyn-tools/blob/master/src/AzureFunctionPackage/Functions/GitHubCreateMergePRs/function.json#L6 (except that azure functions use ncrontab with an additional field - https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-timer#ncrontab-expressions ).  

See - https://crontab.guru/every-12-hours

RikkiGibson:
So the pipeline ran and created #39282 right? Do we need to merge this before the automation will take effect?

Also, you don't have to do this at the same time, but it would be nice if the logs would give the URL of the PRs that it created or found pre-existing.

dibarbet:
> So the pipeline ran and created #39282 right? Do we need to merge this before the automation will take effect?

I've been running the pipeline with dryruns and modified XML configs (to only create a few PRs) on a separate branch manually.  To get this to start going automatically this must be merged to master.

> Also, you don't have to do this at the same time, but it would be nice if the logs would give the URL of the PRs that it created or found pre-existing.

I'll do that in a followup, already got a lot going on here.

dibarbet:
> I've been running the pipeline with dryruns and modified XML configs (to only create a few PRs) on a separate branch manually. To get this to start going automatically this must be merged to master.

Also once this is merged, we'll need to stop running the azure functions.  I believe they can just be deleted?

RikkiGibson:
(Also doesn't need to happen here) I have been thinking that auto-merges should happen several times a day--maybe 4 times--to improve our odds of producing an insertion that passes RPS on a given day, as well as improve our ability to diagnose which code change could have produced an RPS regression when they happen.

/cc @jasonmalinowski on deleting the azure functions. Also @jaredpar might find this work interesting.

jasonmalinowski:
Yes, the azure functions can just be deleted, assuming the auto-merge one is also working again. Don't forget to delete any of the stuff that's involved in deploying it during CI -- @brettfo might be able to help with that.

JoeRobich:
> Also once this is merged, we'll need to stop running the azure functions. I believe they can just be deleted?

@dibarbet I know there is an Azure Function for VSTS Merges. Do we plan to convert that to a pipeline as well?

dibarbet:
> @dibarbet I know there is an Azure Function for creating VSTS Merges. Do we plan to convert that to a pipeline as well?

See the comment from Shyam above - it looks like it's still a WIP and not currently being used.
If it's using the same azure subscription that we need to delete, are we OK to do that and stop publishing it for the time being @shyamnamboodiripad @Cosifne @AbhitejJohn ?  it should be fairly straightforward to convert it to the ado pipeline whenever you're able to start work on it again

Cosifne:
> 
> 
> > @dibarbet I know there is an Azure Function for creating VSTS Merges. Do we plan to convert that to a pipeline as well?
> 
> See the comment from Shyam above - it looks like it's still a WIP and not currently being used.
> If it's using the same azure subscription that we need to delete, are we OK to do that and stop publishing it for the time being @shyamnamboodiripad @Cosifne @AbhitejJohn ? it should be fairly straightforward to convert it to the ado pipeline whenever you're able to start work on it again

I don't think it is in used right now so it should be safe to disable it.

If I remember correctly, the original idea for having that tool is at that time, we want to have a tool to automatically create PR to flow changes from one branch to another. Do we still have that needs right now? @shyamnamboodiripad 
Also, I don't know the reason why that tool is not used right now. The very last time I checked it I think it is because it fails to execute due to some environment set-up problem.

shyamnamboodiripad:
> are we OK to do that and stop publishing it for the time being @shyamnamboodiripad @Cosifne @AbhitejJohn
@dibarbet Based on @Cosifne's reply yes that should be ok.

> Do we still have that needs right now? @shyamnamboodiripad
We don't have an urgent need for it at the moment. But we may need this in the future if want regular code flow between branches (for example, after we snap for a particular release it would be great to have automatic code flow between the snap branch and master so we don't have to do dual checkins all the time). We may also feel need for it once there is parallel work happening in multiple branches in the repo (e.g. 16.4, 16.5 and future)...

brettfo:
You'll need to remove the old Azure function package build/deployment stuff.  See [here](https://github.com/dotnet/roslyn-tools/blob/master/.vsts-ci.yml#L28-L33) and [here](https://github.com/dotnet/roslyn-tools/blob/master/.vsts-ci.yml#L110-L118).

dibarbet:
> You'll need to remove the old Azure function package build/deployment stuff. See [here](https://github.com/dotnet/roslyn-tools/blob/master/.vsts-ci.yml#L28-L33) and [here](https://github.com/dotnet/roslyn-tools/blob/master/.vsts-ci.yml#L110-L118).

Thanks!  I'll go ahead and merge this, disable the azure functions and then send a followup PR to remove the deployment stuff.
I'll leave the project behind with the VSTS though so it can be migrated later.

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -0,0 +1,115 @@
...
+    public static void Main(string[] args)
+    {
+        // Default when no args passed in.
+        var isDryRun = true;
+        var isAutomated = false;
+        var githubToken = string.Empty;
+
+        foreach (var arg in args)
+        {
+            if (arg.StartsWith("--isDryRun"))
```
jaredpar: Nit: should arguments be case sensitive as are done here?

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            else if (arg.StartsWith("--githubToken"))
+            {
+                githubToken = GetArgumentValue(arg);
+            }
+        }
+
+        Console.WriteLine($"Executing with {nameof(isDryRun)}={isDryRun}, {nameof(isAutomated)}={isAutomated}, {nameof(githubToken)}={githubToken}");
+        var config = GetConfig();
+        RunAsync(config, isAutomated, isDryRun, githubToken).GetAwaiter().GetResult();
```
jaredpar: Why not just have `async Main` here so we can properly `await` this call?

src/GitHubCreateMergePRs/Program.cs
dibarbet: I'm not quite understanding - It is case sensitive right now, would it be preferred for them to be case insensitive?

src/GitHubCreateMergePRs/Program.cs
dibarbet: That makes sense!  Will address in followup

PR #563, opened 2019-10-16
Remove azure

dibarbet:
2 items
1.  Remove azure publishing steps.
2.  Fix the scheduled pipeline to not run on every commit.


RikkiGibson:
What's the effect of making the pipeline not run on commit? It feels like if this were a more active repo this could cause breaks to get caught when multiple PRs get merged to master within a short time period.

JoeRobich:
Assuming the code didn't change. It would never generate more than one merge PR per configuration entry. So nothing additional would get caught. Just more work for the pipeline.

PR #564, opened 2019-10-18
Add PRFinder to RoslynTools

RikkiGibson:
/cc @dibarbet @JoeRobich 

this version of it also will fetch and retry if it couldn't find the specified commits (likely, when the commit being inserted was a very recent merge commit to master-vs-deps)

### DIFF
--- a/src/RoslynInsertionTool/PRFinder/Program.cs
+++ b/src/RoslynInsertionTool/PRFinder/Program.cs
@@ -1,0 +1,100 @@
+﻿using LibGit2Sharp;
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Runtime.Serialization.Formatters;
+using System.Text.RegularExpressions;
+
+namespace PRFinder
+{
+    class Program
+    {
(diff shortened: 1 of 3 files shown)


src/RoslynInsertionTool/PRFinder/Program.cs
```diff
@@ -0,0 +1,100 @@
...
+                }
+            }
+
+            return 0;
+        }
+
+        private static void CurrentDomain_UnhandledException(object sender, UnhandledExceptionEventArgs e)
+        {
+            throw new NotImplementedException();
+        }
```
JoeRobich: Is this used?

RikkiGibson: seems not :)

RikkiGibson:
So the plan here is to make PRFinder unnecessary once we get the chance, by fixing RIT to show the changesets properly. I'm going to make some time this afternoon to try and make that happen.

dibarbet:
> So the plan here is to make PRFinder unnecessary once we get the chance, by fixing RIT to show the changesets properly

I'm not sure what the plan is exactly for vs mac insertions, but we'll probably need the same tool when doing those

PR #565, opened 2019-10-19
Make it easier to run the tool locally

RikkiGibson:
This PR adds a `/local` flag to run the tool locally. It prompts you with a nice popup login window. This way you don't have to type your corporate username and password in the command line/save it in VS debug options/other bad stuff like that.

TODO: make pushing the PR branch locally work. I suspect the client credential uses a token but libgit2sharp doesn't expose any way to provide a custom header with the token for a git push. So we would have to either:

1. send a patch to libgit2sharp to add the option (probably not that hard, since they did do the native bridging in libgit2/libgit2sharp#1233)
2. just `Process.Start("git", ...).WaitForExit()`

/cc @JoeRobich @dibarbet @jmarolf

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -16,6 +16,7 @@ using Microsoft.TeamFoundation.Build.WebApi;
 using Microsoft.TeamFoundation.Client;
 using Microsoft.TeamFoundation.Policy.WebApi;
 using Microsoft.TeamFoundation.SourceControl.WebApi;
+using Microsoft.VisualStudio.Services.Client;
 using Microsoft.VisualStudio.Services.Common;
 using Microsoft.VisualStudio.Services.ReleaseManagement.WebApi;
 using Microsoft.VisualStudio.Services.ReleaseManagement.WebApi.Clients;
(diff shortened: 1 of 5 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -407,14 +414,13 @@ private static EnvironmentStatus WaitForReleaseEnvironmentCompletion(string proj
...
-                    Task.Delay(5 * 1000, cancellationToken).Wait();
+                    Task.Delay(5000, cancellationToken).Wait();
 
                     environmentStatus = GetReleaseEnvironment(projectName, release.Id, environmentId, releaseClient, cancellationToken).Status;
 
-                    if(!(environmentStatus.Equals(EnvironmentStatus.InProgress)
-                          || environmentStatus.Equals(EnvironmentStatus.Queued)
-                          || environmentStatus.Equals(EnvironmentStatus.NotStarted)
-                          || environmentStatus.Equals(EnvironmentStatus.Scheduled)))
+                    if (environmentStatus != EnvironmentStatus.InProgress
```
RikkiGibson: this change keeps the tool from spinning for 10 minutes waiting for a release that isn't actually happening.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -296,6 +296,10 @@ public static partial class RoslynInsertionTool
                                 Console.WriteLine("Could not create diff links.");
                             }
 
+                            Console.WriteLine("---PR description---");
+                            Console.WriteLine(prDescription);
```
RikkiGibson: Eventually I'd like to have a dry-run flag for local testing so we can have decent experience finally getting to the bottom of #502

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
dibarbet: Do we need any of this release stuff at all?  What is the release for?

RikkiGibson: The comments make it seem like it is supposed to do symbol publishing for the build we're inserting. But it's not clear to me if that is still something that RIT users need to be worried about, or whether it's something that just happens as part of the signed build pipeline. Can you shed any light @brettfo?

RikkiGibson:
closing in lieu of #570

PR #566, opened 2019-10-22
Add missing equals in xml

dibarbet:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3168484 failed.  Maybe we should add running a dryrun to CI?  Could add a trigger that sets the dry run var to true.
`dotnet run --project .\src\GitHubCreateMergePRs\GitHubCreateMergePRs.csproj` to do a dryrun


brettfo:
My bad, thanks for cleaning up after me!

PR #567, opened 2019-10-22
Trigger dryrun on PR

dibarbet:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3168734

TODO - figure out how to handle secrets from PRs from forks.

### DIFF
--- a/src/GitHubCreateMergePRs/Program.cs
+++ b/src/GitHubCreateMergePRs/Program.cs
@@ -1,12 +1,14 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
+using System.Net.Http.Headers;
 using System.Threading.Tasks;
 using System.Xml.Linq;
 
 public class Program
 {
-    public static void Main(string[] args)
+    public static async Task Main(string[] args)
     {
         // Default when no args passed in.
         var isDryRun = true;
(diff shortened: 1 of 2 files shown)


RikkiGibson:
does dry run need secrets?

dibarbet:
> 
> 
> does dry run need secrets?

github has a much lower rate limit w/out authentication - https://developer.github.com/v3/#rate-limiting
If it doesn't have a secret it will encounter the rate limit.  I'm thinking of just bailing out if we hit a rate limit during a dry run without a secret.  What do you think?

RikkiGibson:
I think give a warning and bail for now when we get rate limited. It feels like some kind of improved flow is possible, such as generating an auth token with a short lifetime and only read permissions per-run. I am definitely no expert in these things though.

dibarbet:
> 
> 
> I think give a warning and bail for now when we get rate limited. It feels like some kind of improved flow is possible, such as generating an auth token with a short lifetime and only read permissions per-run. I am definitely no expert in these things though.

Updated with the bail out piece.

A readonly token that expires often might work, but it could be challenging.  I don't think there is a way in ADO to only share certain secrets with forks, it's all or nothing.  That means we couldn't store the write token and read only token in the variable group as it would have access to both.  

We could potentially have a separate pipeline that runs on PR with dryrun only and then can allow the readonly secret to be used by the fork.

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -74,12 +82,39 @@ private static XDocument GetConfig()
...
             Console.WriteLine($"##vso[task.logissue type=error]Error creating PR. GH response code: {error.StatusCode}");
             Console.WriteLine($"##vso[task.logissue type=error]{await error.Content.ReadAsStringAsync()}");
+
+            // Github rate limits are much lower for unauthenticated users.  We will definitely hit a rate limit
+            // If we hit it during a dryrun, just bail out.
+            if (isDryRun && string.IsNullOrWhiteSpace(githubToken))
+            {
+                if (TryGetRemainingRateLimit(error.Headers, out var remainingRateLimit) && remainingRateLimit == 0)
+                {
+                    Console.WriteLine($"##vso[task.logissue type=error]Hit GitHub rate limit in dryrun with no auth token.  Bailing out.");
```
RikkiGibson: Do we have a run where this line has been hit?

dibarbet: In the PR checks :)
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3169149

PR #568, opened 2019-10-23
Snap for 16.4 P3

dibarbet:
Draft - branches are not created yet.

### DIFF
--- a/src/GitHubCreateMergePRs/config.xml
+++ b/src/GitHubCreateMergePRs/config.xml
@@ -6,8 +6,7 @@
     <merge from="dev16.0" to="release/dev16.1" />
     <merge from="release/dev16.1" to="release/dev16.2" />
     <merge from="release/dev16.2" to="release/dev16.3" />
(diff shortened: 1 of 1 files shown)


src/GitHubCreateMergePRs/config.xml
```diff
@@ -8,8 +8,8 @@
     <merge from="release/dev16.2" to="release/dev16.3" />
     <merge from="release/dev16.3" to="release/dev16.4-preview1" />
```
RikkiGibson: are we ready to delete preview1?

dibarbet: I wasn't sure - what's the policy on servicing past previews?

RikkiGibson: I think when we close out a preview branch from QB, we keep it around for one cycle just in case there's some emergency servicing thing and then get rid of it

dibarbet: I checked with Jason as well, since preview 2 is shipped we can delete

PR #573, opened 2019-10-24
Fix change log

### LINKED ISSUE #502: RIT diff links and commit histories are not working

### LINKED ISSUE #550: When the message on description contains a PR reference it should link to it
https://github.com/dotnet/roslyn-tools/blob/eb65ecf22007c3627913a6b308b549f6a47729a9/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs#L643

Currently when creating the message if it says something along the lines of "Merge pull request #1234 from user/branch" The number creates an automatic link to https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1234 instead of going to https://github.com/{organization}/{repo}/pull/1234

RikkiGibson:
Closes #502
Closes #550

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "2.1.401"
+    "dotnet": "3.0.100"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.18473.8"
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.0.100"
+    "dotnet": "2.1.401"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.18473.8"
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -535,7 +535,6 @@ namespace Roslyn.Insertion
         private static readonly Regex IsMergePRCommit = new Regex(@"^Merge pull request #(\d+) from");
         private static readonly Regex IsSquashedPRCommit = new Regex(@"\(#(\d+)\)$");
 
-#nullable enable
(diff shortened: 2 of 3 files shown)


RikkiGibson:
```cs
RoslynInsertionTool.VisualStudioTeamServices.cs(538,2): error CS1024: Preprocessor directive expected [D:\a\1\s\src\RoslynInsertionTool\RoslynInsertionTool\RoslynInsertionTool.csproj]
RoslynInsertionTool.VisualStudioTeamServices.cs(588,2): error CS1024: Preprocessor directive expected [D:\a\1\s\src\RoslynInsertionTool\RoslynInsertionTool\RoslynInsertionTool.csproj]
```

Looks like we need to update the compiler used in vsts

jmarolf:
This is because we are building with the 2.1 SDK not the 3.0 SDK

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "2.1.401"
+    "dotnet": "3.0.100"
```
dibarbet: can you update the version in here too to match?
https://github.com/dotnet/roslyn-tools/blob/master/merge-pr-pipeline.yml

There might be a way to pull the version from that file, not sure though.

dibarbet:
To get the dotnet bot auth token (if you think it's appropriate to use that token)
It's already attached to a variable group - https://devdiv.visualstudio.com/DevDiv/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=234&path=RoslynMergeBot

In the pipeline you have to link the variable group (see variables tab), then pass it into the parameters to the one off insertion script in the tasks tab in the pipeline, something like `$(dotnet-bot-github-auth-token)`.  Of course the one off insertion script has to pass it along as well.

At some point it might also just be worth actually defining the insertion pipeline in this repo, instead of just nebulously out there...

jmarolf:
@RikkiGibson 	I'm just going to back out the nullable change and have that be in a separate PR. The infra for this repo is _way_ too out of date for this change to be made cleanly at this time (looks like no one ever updated this repo to receive arcade updates)

RikkiGibson:
seems like i had backed out of passing a nullable parameter to the method anyway!

RikkiGibson:
@dibarbet I don't want to derail any work you're doing related to snap so I will wait for your go-ahead.

dibarbet:
You're good to go!   Thanks for waiting.

PR #574, opened 2019-10-24
update arcade

jmarolf:


jmarolf:
This cannot move forward. Arcade broke generating a nuget package from a nuspec in [this](https://github.com/dotnet/arcade/pull/2003) PR.  We will need to submit a PR to arcade to do the correct thing in the presence of nuspec files.  The root of the issue is that is appends the `CommonMetadataElements` xml in to the `NuspecProperties` list. The pack command does not know how to extract key-value pairs from xml

Issue #572, opened 2019-10-23
Don't download entire VSSetup artifacts


RikkiGibson:
When the tool runs normally it downloads the Roslyn-Signed build VSSetup artifacts. This is about 200 MB. But it then just looks at a versions csv file and a few other small things. It feels like more granular APIs could be used here to bring the download size way down.

RikkiGibson:
should be possible using a newer ADO client library and this API: https://docs.microsoft.com/en-us/dotnet/api/microsoft.teamfoundation.build.webapi.buildhttpclientbase.getfilecontentsasync?view=azure-devops-dotnet#Microsoft_TeamFoundation_Build_WebApi_BuildHttpClientBase_GetFileContentsAsync_System_String_System_String_System_Nullable_System_Guid__System_String_System_String_System_String_System_Object_System_Threading_CancellationToken_

RikkiGibson:
This will likely save less than 1 minute in runtime.

PR #575, opened 2019-10-24
Don't use local VS enlistment to create insertions

### LINKED ISSUE #571: Don't clone VS to create insertions
It seems like Azure DevOps has the API necessary to avoid cloning VS. We should do this so our infrastructure runs faster.

https://github.com/microsoft/azure-devops-dotnet-samples/blob/master/ClientLibrary/Samples/Git/PullRequestsSample.cs#L98

By doing this we might actually end up getting rid of the dependency on libgit2sharp, which would allow us to start using simpler authentication like in #565.

RikkiGibson:
Closes #571


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.csproj
```diff
@@ -13,7 +13,6 @@
     <Reference Include="System.Xml.Linq" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="LibGit2Sharp" Version="0.22.0" />
```
jmarolf: 💯 🥇 🎆 🍾

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -256,10 +254,14 @@ private static void PopulateComponentToFileMapForFile(JObject jDocument, string
...
-        private static JObject GetJsonDocumentForComponentsFile(string componentsJSONPath)
+        private static JObject GetJsonDocumentForComponentsFile(
+            GitHttpClient gitClient,
+            RoslynInsertionToolOptions options,
+            string componentsJSONPath)
         {
             JObject jsonDocument = null;
 
+            // TODO: use gitClient
             if (File.Exists(componentsJSONPath))
```
dibarbet: just curious - what is this stuff doing and why do we not care about it yet (I'm presuming the file will never exist now)

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -380,20 +351,20 @@ public static partial class RoslynInsertionTool
...
-            cancellationToken.ThrowIfCancellationRequested();
-            Console.WriteLine("Updating CoreXT config file");
-            coreXT.SaveConfig();
+        public static string ToFullString(this XDocument document)
+        {
+            return document.Declaration.ToString() + Environment.NewLine + document.ToString();
         }
 
+        // TODO: either delete or actually use this
         private static string GetHTMLSuccessMessage(GitPullRequest pullRequest, List<string> newPackageFiles)
```
jmarolf: Just delete this message.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -221,88 +214,95 @@ public static partial class RoslynInsertionTool
                 // ********************* Verify Build Completes **************************
                 if (Options.PartitionsToBuild != null)
                 {
-                    Console.WriteLine($"Verifying build succeeds with changes");
-                    foreach (var partition in Options.PartitionsToBuild)
-                    {
-                        Console.WriteLine($"Starting build of {partition}");
+                    // TODO: remove PartitionsToBuild command line option
```
jmarolf: I am fine with this being in a different PR, but we should remove it

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -221,88 +214,95 @@ public static partial class RoslynInsertionTool
...
+                    Comment = $"Updating {Options.InsertionName} to {buildVersion}",
+                    Changes = allChanges
+                };
+                var push = new GitPush
+                {
+                    RefUpdates = new[] { newBranch },
+                    Commits = new[] { commit }
+                };
+
+                // TODO: do we need to specify --force? how can we do that?
```
jmarolf: You do not need to specify --force

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -221,88 +214,95 @@ public static partial class RoslynInsertionTool
...
+                    // TODO: remove PartitionsToBuild command line option
+                }
 
-                        if (!(await CanBuildPartitionAsync(partition, cancellationToken)))
-                        {
-                            Console.WriteLine($"Build of partition {partition} failed");
-                            return (false, 0);
-                        }
+                // ********************* Create push *************************************
+                var gitClient = ProjectCollection.GetClient<GitHttpClient>();
```
jmarolf: I would move this to its own method

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -221,88 +214,95 @@ public static partial class RoslynInsertionTool
...
+                var pullRequestId = Options.UpdateExistingPr;
+                var useExistingPr = pullRequestId != 0;
+
+                var prDescription = $"Updating {Options.InsertionName} to {buildVersion} ([{commitSHA}]({lastCommitUrl}))";
+                try
                 {
-                    cancellationToken.ThrowIfCancellationRequested();
-                    var prDescription = $"Updating {Options.InsertionName} to {buildVersion} ([{commitSHA}]({lastCommitUrl}))";
-                    if (useExistingPr && pullRequest != null)
+                    // TODO: maybe skip if useExistingPR && !Options.OverwritePR
```
jmarolf: yes we should skip it

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -5,97 +5,96 @@
...
-        public string ConfigFilePath { get; }
-        public string ComponentsFilePath { get; }
-        public XDocument Config { get; }
         private static Dictionary<string, string> ComponentToFileMap;
         private static Dictionary<string, JObject> ComponentFileToDocumentMap;
         private static HashSet<string> dirtyFiles;
 
-        public CoreXT(string configPath, XDocument config, string componentsJSONPath)
+        private const string DefaultConfigPath = ".corext/Configs/default.config";
+        private const string ComponentsJsonPath = ".corext/Configs/components.json";
```
jmarolf: roslyn insertion components do not live in components.json, you will need to examine every json file that lives in `.corext/Configs`

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -127,7 +126,7 @@ internal void AddNewPackage(PackageInfo packageInfo)
         public void UpdatePackageVersion(PackageInfo packageInfo)
         {
             var versionAttribute = GetVersionAttribute(packageInfo);
-            var currentVersion = GetPackageVersion(versionAttribute);
+            var currentVersion = GetPackageVersion(versionAttribute); // TODO: remove dead local
```
jmarolf: do it. what are you? scared?

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -218,8 +216,8 @@ private static void PopulateComponentJsonMaps(string mainComponentsJsonPath, str
 
                         if (!string.IsNullOrEmpty(subComponentFileName))
                         {
-                            var componentsJSONPath = Path.Combine(enlistmentRoot, ".corext", "Configs", subComponentFileName);
-                            var jDoc = GetJsonDocumentForComponentsFile(componentsJSONPath);
+                            var componentsJSONPath = Path.Combine(".corext", "Configs", subComponentFileName);
```
jmarolf: ah disregard my previousl comment I see you handle this correctly here.

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
RikkiGibson: I believe what it's actually doing is downloading VSSetup for the build we're inserting (about 200MB) and reading a few config files from it. see #572

jmarolf: This whole method can be removed. we can just look up the version file from the build that we find

dibarbet:
also would generally suggest running a test build on another repo that uses rit?  Ex unit testing just for sanity

jmarolf:
> also would generally suggest running a test build on another repo that uses rit? Ex unit testing just for sanity

or project system which inserts more than all the other repos combined :)

RikkiGibson:
From adhoc testing this seems to reduce insertion times from ~40 minutes to ~5 minutes.

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
RikkiGibson: Pushing this off to a future PR

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -313,14 +313,10 @@ private static async Task<string> DownloadArtifactsAsync(BuildHttpClient buildCl
 
             using (Stream s = await buildClient.GetArtifactContentZipAsync(Options.TFSProjectName, build.Id, artifact.Name, cancellationToken))
             {
-                using (var fs = File.OpenWrite(archiveDownloadPath))
+                using (ZipArchive archive = new ZipArchive(s))
                 {
-                    // Using the default buffer size.
-                    await s.CopyToAsync(fs, 81920, cancellationToken);
+                    archive.ExtractToDirectory(tempDirectory);
```
RikkiGibson: this change seems to make things a bit faster. instead of saving the whole zipfile to disk, we just stream the zip file into memory and then extract it to the destination folder.

RikkiGibson:
It seems like the tool is creating good insertions now. It changes some whitespace in a few files but it feels like the easiest thing is just to let it do the change. Then it shouldn't try to change the whitespace again in the future.

Will try and merge and update the pipelines on Monday, barring any other feedback.

RikkiGibson:
I confirmed that this produces good insertions on the existing (slow) pipelines, so I will merge now.

PR #584, opened 2019-10-31
Don't update Stub.exe.config or versions.xml

RikkiGibson:
/cc @tmat

related to https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1011582


RikkiGibson:
This seemed to work as expected in local testing, so we can just merge whenever we decide we're sure that this should go in.

PR #588, opened 2019-11-06
UpdateExistingPr should get file versions from PR branch

### LINKED ISSUE #587: UpdateExistingPr uses file version from base branch instead of from PR
Essentially this check happens later than it should, so we end up using and modifying file versions from the base branch, overwriting other changes to those files in the existing PR:

https://github.com/dotnet/roslyn-tools/blob/8608f6754b869da20cc6cae9ecbb033d2025f65e/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs#L257

/cc @shyamnamboodiripad @AbhitejJohn

RikkiGibson:
Closes #587 

Also makes the behavior around failing to get the old build a little more graceful. ~~When the old build is not from the same branch we fail to look it up. This can be fixed, just want to do it separately.~~ Ended up just doing it, the REST API to get builds has a buildNumber parameter so it's pretty simple.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -194,10 +194,16 @@ namespace Roslyn.Insertion
         private static async Task<Build> GetSpecificBuildAsync(BuildVersion version, CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
-            Console.WriteLine($"Getting latest passing build from {Options.TFSProjectName} where name is {Options.BuildQueueName}");
+            Console.WriteLine($"Getting build with build number {version}");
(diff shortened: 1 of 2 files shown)


RikkiGibson:
I manually created some dual insertions with this, it seems like it works as expected now. @shyamnamboodiripad @AbhitejJohn

shyamnamboodiripad:
Thanks for fixing @RikkiGibson! We'll keep an eye out if we see any other weirdness.

We were also running into some weirdness around semantic versioning of our nuget packages that @AbhitejJohn was looking into. He can probably comment on the finding better. But looks like insertion tool was considering a *.11 version to be lower than *.2 and refusing to update default.config as a result. I don't think we would hit this normally as we usually don't have 11 builds happening in 1 day :) - but it happened yesterday as we merged a bunch of changes that were all awaiging QB approval...

@AbhitejJohn can probably comment better on what he found here.

RikkiGibson:
Weird! Would be good to create a new issue for this.

PR #589, opened 2019-11-13
Update arcade

### LINKED ISSUE #577: Update .NET SDK version
This requires fixing an Arcade bug which breaks publishing RIT. /cc @jmarolf

https://github.com/dotnet/arcade/pull/2003/files#diff-ca11509da9be27e6202dd06024395ed4

After this we should be able to start regularly consuming Arcade updates and that way remain up to date.

RikkiGibson:
Closes #577 
Continuation of #574 
Related to dotnet/arcade#4341


global.json
```diff
@@ -1,8 +1,13 @@
 {
   "tools": {
-    "dotnet": "2.1.401"
+    "dotnet": "3.0.100",
+    "runtimes": {
+      "dotnet": [
+        "2.1.11"
```
JoeRobich: Is this for the xunit test runner?

RikkiGibson: Actually good question. @jmarolf could we maybe just delete this and automatically use the runtime included with 3.0.100 SDK?

jmarolf: yep

RikkiGibson: Will probably do this in a future PR

Issue #621, opened 2019-12-12
Auto-PR bot: improve experience around merge conflicts


RikkiGibson:
When an auto-PR has merge conflicts the only way to find out about it is to check in the evening or the next morning and see that for some reason an auto-PR has still not been merged. The bot should take a few automatic steps to triage PRs with merge conflicts, potentially including:

1. Performing a merge locally as we have seen some merges fix themselves up when using the local git merge algorithm as opposed to the GitHub one
2. Performing a merge using the [Merge API](https://developer.github.com/v3/repos/merging/) and giving a notification to the infraswat team when it fails

/cc @dibarbet @JoeRobich

JoeRobich:
Closing as the GH merge bot is nearing retirement

PR #624, opened 2019-12-13
Add extra label to PRs with merge conflicts

RikkiGibson:
Related to #621 

At the time we create the merge PR, we notice if there are conflicts and add an extra label if that is the case. /cc @dibarbet 

This handles some scenarios but not all. Some unhandled scenarios include:
1. The bot creates merges for master-to-master-vs-deps and 16.5-preview1-vs-deps to master-vs-deps. Once one is merged, we can start to have conflicts in the other merge.
2. The bot creates a merge for master-to-master-vs-deps. Then someone merges a PR to one of the branches which makes the bot PR start to have merge conflicts.

GitHub really should just show if a PR is known to have merge conflicts in the PR list just as it shows the checks results. But this is an inexpensive stopgap that gets us a lot of the triage improvements that we do want.

Example PR: rikkigibson/roslyn#6

### DIFF
--- a/src/GithubMergeTool/GithubMergeTool.cs
+++ b/src/GithubMergeTool/GithubMergeTool.cs
@@ -198,7 +198,13 @@ Once all conflicts are resolved and all the tests pass, you are free to merge th
                     jsonBody = JObject.Parse(await response.Content.ReadAsStringAsync());
                     mergeable = (bool?)jsonBody["mergeable"];
                 }
+
                 Console.WriteLine();
+
+                if (attempt == maxAttempts)
+                {
+                    Console.WriteLine($"##vso[task.logissue type=warning]Timed out waiting for PR mergeability status to become available.");
+                }
             }
 
 
--- a/src/GithubMergeTool/GithubMergeTool.cs
+++ b/src/GithubMergeTool/GithubMergeTool.cs
@@ -1,11 +1,13 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Text;
 using System.Threading.Tasks;
+using Newtonsoft.Json;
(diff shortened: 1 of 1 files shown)


src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -180,9 +182,40 @@ git commit
             jsonBody = JObject.Parse(await response.Content.ReadAsStringAsync());
 
             var prNumber = (string)jsonBody["number"];
+            var mergeable = (bool?)jsonBody["mergeable"];
+            if (mergeable == null)
```
dibarbet: :( is the mergeable attribute computed asynchronously?

src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -180,9 +182,40 @@ git commit
...
+            {
+                const int maxAttempts = 5;
+                var attempt = 0;
+
+                Console.Write("Waiting for mergeable status");
+                while (mergeable == null && attempt < maxAttempts)
```
dibarbet: consider logging an error/warning if attempt exceeds maxAttempts.  I would guess the mergeable attribute should always eventually be computed (either false or true).

src/GithubMergeTool/GithubMergeTool.cs
RikkiGibson: correct, it takes two seconds or so for dotnet/roslyn, but probably depends on complexity of changes

src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -180,9 +182,46 @@ git commit
...
+                    Console.Write(".");
+                    await Task.Delay(1000);
+                    response = await _client.GetAsync($"repos/{repoOwner}/{repoName}/pulls/{prNumber}");
+                    jsonBody = JObject.Parse(await response.Content.ReadAsStringAsync());
+                    mergeable = (bool?)jsonBody["mergeable"];
+                }
+
+                Console.WriteLine();
+
+                if (attempt == maxAttempts)
```
RikkiGibson: this should probably check `&& mergeable is null`

PR #670, opened 2020-01-24
Run the merge pipeline every 3 hours

dibarbet:
Trying to gauge thoughts on running merges every three hours instead of every 12.  

We specifically want to do this for Roslyn -vs-deps branches, but if an overall change to 3 works then we can do it this way rather than more difficult changes

Questions -
1.  Do we need to have different schedules for different repos?
2.  Should we only run on a more frequent schedule for specific branches?

If the answer to either 1 or 2 is Yes, then we'll need to make more invasive changes (e.g. multiple pipelines)

 @jmarolf @dmonroym @brettfo @mavasani @JoeRobich for thoughts

### DIFF
--- a/merge-pr-pipeline.yml
+++ b/merge-pr-pipeline.yml
@@ -16,7 +16,7 @@ pr:
 - master
 
 schedules:
-- cron: "0 */12 * * *"
+- cron: "0 */3 * * *"
   displayName: Roslyn Merge Tool
   branches:
     include:


brettfo:
F# would be interested in being able to specify daily runs, if possible.  It'll be another work item, but perhaps a `cadence="(daily|twice-daily|etc)"` or similar attribute on the `<repo>` node would help.  Short term if Roslyn needs every 3 hours, we can deal with that so that others aren't blocked waiting for a merge.

dibarbet:
going to merge this, let me know if anyone encounters problems with the frequency.

JoeRobich:
We support a weekly frequency (https://github.com/dotnet/roslyn-tools/blob/master/src/GitHubCreateMergePRs/Program.cs#L127), daily wouldn't be hard to implement if needed.

RikkiGibson:
It can be hard to do correctly but the cost of messing it up is probably low. For instance the implementation we have now will probably produce insertions every 3 hours for "weekly" repos, just only on Sundays.

JoeRobich:
> It can be hard to do correctly but the cost of messing it up is probably low. For instance the implementation we have now will probably produce insertions every 3 hours for "weekly" repos, just only on Sundays.

Sounds like we should check if the time is roughly 00:00 UTC =p

Issue #666, opened 2020-01-22
Insertion tool to pick up partially successful builds as well.


AbhitejJohn:
Currently the Roslyn Insertion Tool automatically picks up the latest build from an AzDO definition for insertion into the VS repo. This filters down to builds that have Succeeded only. Can we alter this to include Partially Succeeded builds as well? 

This is tied to the Unit Testing definition that sometimes ends up in this state cause a source Indexing task flakes out. See [this ](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3404317&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=ee28fa1a-e4be-5688-e7d9-bc085d0d4fdd&l=18)for instance. Unfortunately there isn't an easy way to fix the indexing job cause its tied to machine infra. 

/cc: @RikkiGibson

RikkiGibson:
If we find that all the "Partially Succeeded" builds for the components we insert are actually OK to insert, then let's just change the constant value passed here:

https://github.com/dotnet/roslyn-tools/blob/09f1886016ffa152104ec6b3f5e57bfb5c669127/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs#L79

If there are actually cases where we don't want to insert a partially succeeded build (even if it contains all the right artifacts?), then let's add a command line argument that just your insertion release can run.

/cc @JoeRobich

RikkiGibson:
Looked a bit and I only see some Project System builds which seem to be Partially Succeeded for the same reason as Unit Testing. Tagging @dmonroym in case you have any concerns about changing this.

I'd personally be fine with making the tool attempt to pick up Partially Succeeded builds, but it should log a warning in the release run log (similar to how other warnings are logged now) and it should maybe add a note in the PR description that it's using a Partially Succeeded build.

AbhitejJohn:
Thanks Rikki. Yup a commandline switch should be alright too.

dmonroym:
Don't see a problem with this for now.

dmonroym:
Closing this as PR has been merged and everything seems to be running smoothly still.
@AbhitejJohn

PR #674, opened 2020-01-27
Fix to insert partially successful builds

AbhitejJohn:
Fix to pick up latest builds that have partially succeeded as well with proper warnings. We dont anticipate this to happen unless there is a non-critical tooling problem like failing to index sources.

Addresses #666 .


AbhitejJohn:
Here's a test insertion: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/225173?_a=overview

PR #677, opened 2020-01-29
Add option to update an existing PR if it's already exist

genlu:
FYI @dibarbet @JoeRobich @RikkiGibson 

I was a little concerned about the case that it triggers a rebuild when existing run is almost finished, thus delay our codeflow. But looking at the time from PR creation to auto-merge in Roslyn, it usually happens in about 1.5 hours, well within the 3 hours window. So in most cases, we will only reset when CI failed or stuck (due to infra issue), which is actually desired.

Another change I'm considering is to rerun CI if it hasn't finished in certain time, which hopefully can mitigate some transient infra issue. Thoughts?

### DIFF
--- a/src/GitHubCreateMergePRs/Program.cs
+++ b/src/GitHubCreateMergePRs/Program.cs
@@ -38,7 +38,7 @@ public class Program
 
         Console.WriteLine($"Executing with {nameof(updateExistingPr)}={updateExistingPr}, {nameof(isDryRun)}={isDryRun}, {nameof(isAutomated)}={isAutomated}, {nameof(githubToken)}={githubToken}");
         var config = GetConfig();
-        await RunAsync(config, updateExistingPr, isAutomated, isDryRun, githubToken);
+        await RunAsync(config, isAutomated, isDryRun, githubToken);
     }
 
     private static string GetArgumentValue(string arg)
@@ -118,13 +118,17 @@ public class Program
         return false;
     }
 
-    private static async Task RunAsync(XDocument config, bool updateExistingPr, bool isAutomatedRun, bool isDryRun, string githubToken)
+    private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool isDryRun, string githubToken)
     {
         var gh = new GithubMergeTool.GithubMergeTool("dotnet-bot@users.noreply.github.com", githubToken, isDryRun);
         foreach (var repo in config.Root.Elements("repo"))
         {
             var owner = repo.Attribute("owner").Value;
             var name = repo.Attribute("name").Value;
+
+            // We don't try to update existing PR unless asked.
+            var updateExistingPr = bool.Parse(repo.Attribute("updateExistingPr")?.Value ?? "false");
+
             foreach (var merge in repo.Elements("merge"))
             {
                 var fromBranch = merge.Attribute("from").Value;
--- a/src/GitHubCreateMergePRs/Program.cs
+++ b/src/GitHubCreateMergePRs/Program.cs
@@ -13,6 +13,7 @@ public class Program
         // Default when no args passed in.
         var isDryRun = true;
         var isAutomated = false;
+        var updateExistingPr = true;
         var githubToken = string.Empty;
 
         foreach (var arg in args)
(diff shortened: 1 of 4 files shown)


genlu:
@JoeRobich Who else is using the merge tool? Probably need to let them know about this change before merge.

JoeRobich:
from the [config ](https://github.com/dotnet/roslyn-tools/blob/master/src/GitHubCreateMergePRs/config.xml) it looks like project system, f#, test...

genlu:
FYI @AbhitejJohn @mavasani @brettfo @jmarolf 
Do you have any concerns about this change?

RikkiGibson:
If anyone is using the tool with CI runs that take 3+ hours they might be unhappy. If this is the case you could just make the tool check the `<repo>` element for an optional attribute that configures this behavior.

brettfo:
How often does the tool run?  I'm concerned that I might be fixing merge conflicts/tests/etc. and the tool will re-run and overwrite my changes.  Or the tool could run a few minutes after I've pushed fixes and it'll reset the merge discarding my fixes.

brettfo:
I'd prefer to keep the behavior as-is and for teams that want an updated merge, they can delete the stalled/bad/etc. merge and a new one will be created next run.

JoeRobich:
Can we use the `merge conflicts` tag to skip updating PRs with conflicts?

brettfo:
That still may not be sufficient; I might have test failures when merging feature branches.  This isn't uncommon in F#.

genlu:
> How often does the tool run? 

Every 3 hours now

> make the tool check the <repo> element for an optional attribute

I like this idea, will add and enable only for Roslyn

> Can we use the merge conflicts tag to skip updating PRs with conflicts?

Like this too, will add.

genlu:
I have pushed change as mentioned above. now we don't reset branch if:
1. There's "Merge Conflicts" label, or
2. "updateExistingPr" isn't set to true for the repo in the config file.

Could someone please explain how the "Merge Conflicts" label works? So once the conflict is resolved, is it supposed to be removed? By whom?

src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -97,13 +93,57 @@ public class GithubMergeTool
...
+            var existingPR = JArray.Parse(await prsResponse.Content.ReadAsStringAsync()).FirstOrDefault(pr => (string)pr["title"] == prTitle);
+            if (existingPR != null)
+            {
+                if (updateExistingPr)
+                {
+                    // Get the SHA of the PR branch HEAD
+                    var prSha = ((JValue)existingPR["head"]["sha"]).ToObject<string>();
+                    var existingPrNumber = ((JValue)existingPR["number"]).ToObject<string>();
+
+                    if (prSha != srcSha)
```
RikkiGibson: Does this assume that if the source branch is on a different SHA than the merge branch, there are new commits in the source branch? If you pushed a commit to resolve merge conflicts or test failures then would this code throw that away?

Perhaps if "force" were disabled then we would gracefully fail to update the merge PR in the case that new commits were added to it directly?

RikkiGibson:
The Merge Conflicts label is pretty primitive. I implemented it in #624. At the time of PR creation, it just waits a few seconds for GitHub to compute the mergability status. Merge conflicts can still be introduced later on (e.g. from the base branch getting updated after the creation of the merge PR). The tool will not catch that.

Have you tried running the

genlu:
> Have you tried running the

@RikkiGibson  Running what ? :)

src/GithubMergeTool/GithubMergeTool.cs
genlu: I'm not sure what we will end up with in the PR by using `reset --mixed`. We can do a `git merge --ff-only` instead, which would be a no-op if there's conflict resolving commit in the merge branch.

src/GithubMergeTool/GithubMergeTool.cs
```diff
@@ -97,13 +93,54 @@ public class GithubMergeTool
...
+                        await PostComment(existingPrNumber, $"Reset HEAD of `{prBranchName}` to `{srcSha}`");
+
+                        // Check for merge conflicts again after reset.
+                        existingPrMergeable = await IsPrMergeable(existingPrNumber);
+                    }
+
+                    // Add label if there's merge conflicts even if we made no change to merge branch,
+                    // since can also be introduced by change in destination branch. It's no-op if the
+                    // label already exists. 
+                    if (existingPrMergeable == false)
```
genlu: @RikkiGibson We are now checking mergeable state directly, and add "merge conflicts" label accordingly during update too.

RikkiGibson:
> Running what ? :)

🤦‍♂ I have no idea what the end of that sentence was. Sorry..

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -13,6 +13,7 @@ public static async Task Main(string[] args)
         // Default when no args passed in.
         var isDryRun = true;
         var isAutomated = false;
+        var updateExistingPr = true;
```
RikkiGibson: Should this default to 'false'?

RikkiGibson: Actually, for simplicity it feels like this shouldn't be a command line arg, it should only be specified in config.xml per-repo. Do you agree?

genlu: Sure, sounds reasonable.

RikkiGibson:
Have you created a test PR using your version of the tool? One easy way to do this is to go in config.xml locally and change `<repo owner="dotnet" name="roslyn" ...` to your fork, and comment out other repos. With and without the new config setting to see if it behaves as expected (no need to wait 3 hours :smile: just run it consecutively with different configurations)

If there are no branches to merge in your fork you can also create dummy branches and have the config reference them.

genlu:
@RikkiGibson I think I have addressed your prior comments, any more concerns?

PR #687, opened 2020-02-04
Add merges for 16.5-preview3

agocke:


src/GitHubCreateMergePRs/config.xml
```diff
@@ -6,7 +6,8 @@
     <merge from="dev16.0" to="release/dev16.3" />
     <merge from="release/dev16.3" to="release/dev16.4" />
     <merge from="release/dev16.4" to="release/dev16.5-preview2" />
```
dibarbet: do we need the preview2 branch anymore?  Doesn't seem likely we'll need to service it since we're so close to p3

PR #691, opened 2020-02-04
Fix json

genlu:
Fix error:

```
##[error]Error creating PR. GH response code: UnprocessableEntity
##[error]{"message":"Invalid request.\n\nFor 'properties/force', \"true\" is not a boolean.","documentation_url":"https://developer.github.com/v3"}
```

### DIFF
--- a/src/GithubMergeTool/GithubMergeTool.cs
+++ b/src/GithubMergeTool/GithubMergeTool.cs
@@ -1,4 +1,6 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using System;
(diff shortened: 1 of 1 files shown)


RikkiGibson:
I think it's time we stop rolling our own JSON strings and start writing a few POCOs or even anonymous objects to [de]serialize with newtonsoft.

You don't need to do that here and now but it really seems like it would have a positive impact on readability and correctness (and a small impact on cruft/boilerplate).

RikkiGibson:
e.g.

https://github.com/dotnet/roslyn-tools/blob/f0f99c9278c8bb8ad1a37de5e9b3c971dfa8c825/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs#L441

genlu:
@RikkiGibson Agreed. I'm planning to take a look at this once this simple fix is merge.

PR #709, opened 2020-02-14
Cleaning up a stale script

AbhitejJohn:
Unit Testing has moved on to use OneOffInsertion as everyone else. Doesn't look like we need this anymore unless someone else is using this.


RikkiGibson:
Make sure also to delete or disable the pipeline :)

AbhitejJohn:
@RikkiGibson : already gone. 👍

PR #710, opened 2020-02-14
Update Roslyn code flow configuration

JoeRobich:
We should come up with a plan for getting our service branches consistent.

### DIFF
--- a/src/GitHubCreateMergePRs/config.xml
+++ b/src/GitHubCreateMergePRs/config.xml
@@ -1,22 +1,20 @@
 <config>
   <repo owner="dotnet" name="roslyn" updateExistingPr="true">
+    <!-- Roslyn servicing branches -->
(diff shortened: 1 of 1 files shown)


src/GitHubCreateMergePRs/config.xml
```diff
@@ -1,22 +1,20 @@
 <config>
   <repo owner="dotnet" name="roslyn" updateExistingPr="true">
+    <!-- Roslyn servicing branches -->
+    <merge from="dev15.9.x" to="dev16.0-vs-deps" />
+    <merge from="dev16.0-vs-deps" to="release/dev16.3" />
```
RikkiGibson: what is dev16.0-vs-deps? it seems like that branch shouldn't exist if we have a requirement for our contributors to use dev16.

JoeRobich: There is no dev16.0 branch anymore

PR #722, opened 2020-02-24
Use NuGet.Versioning for SemanticVersion

### LINKED ISSUE #719: RIT does not compare SemVer correctly
Roslyn changed from using `beta1` as the prerelease label to using `1`. Our version numbers changed from the form '3.5.0-beta4-20121-02' to the form '3.6.0-1.20121.8'.

I noticed that once the build number for the day got into the double digits that it no longer updated the Roslyn version being inserted. Looking at the RIT log. 

With version '3.6.0-1.20121.8' being inserted the following log was produced.
```
2020-02-21T22:03:58.6176287Z Processing package 'Roslyn-Signed/VSSetup\DevDivPackages\VS.ExternalAPIs.Roslyn.3.6.0-1.20121.12.nupkg'
2020-02-21T22:03:58.6176882Z Package 'VS.ExternalAPIs.Roslyn.3.6.0-1.20121.12' doesn't need to be inserted, version is lower than or equal to the one already inserted.
```

JoeRobich:
Running the following gives `3.6.0-1.20121.8 > 3.6.0-1.20121.12: False`
```csharp
using System;
using NuGet.Versioning;

namespace ConsoleApp12
{
    class Program
    {
        static void Main(string[] args)
        {
            SemanticVersion version1 = SemanticVersion.Parse("3.6.0-1.20121.8");
            SemanticVersion version2 = SemanticVersion.Parse("3.6.0-1.20121.12");

            Console.WriteLine($"{version1} > {version2}: {version1 > version2}");
        }
    }
}
```
Fixes #719


src/RoslynInsertionTool/RoslynInsertionTool.Commandline/RIT.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <RootNamespace>RoslynInsertionTool</RootNamespace>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net46</TargetFramework>
+    <TargetFramework>net472</TargetFramework>
```
RikkiGibson: Pipelines may need to be manually updated after this change. Check the OneOffInsertion step in https://devdiv.visualstudio.com/DevDiv/_releaseDefinition?definitionId=2114&_a=definition-tasks&environmentId=11199

Ideally, we could also change packaging to also put the tool executables and scripts in some simpler path that won't change when we change the TFM.

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.TeamFoundation.SourceControl.WebApi;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using NuGet.Versioning;
```
RikkiGibson: why do these files only contain using changes? is that enough to just change what things bind to in the existing code and it all just works?

JoeRobich: Yup. Practically a drop in replacement.

src/RoslynInsertionTool/RoslynInsertionTool.Commandline/RIT.csproj
JoeRobich: Downgraded to v4.9.4 so we can keep net46 tfm for now

PR #723, opened 2020-02-24
Ran 'dotnet format -f .'

JoeRobich:


src/BuildRetainer/Mono.Options/Options.cs
```diff
@@ -153,1214 +153,1311 @@ namespace NDesk.Options
 namespace Mono.Options
 #endif
 {
-	static class StringCoda {
```
RikkiGibson: I do wonder if this BuildRetainer project is used any more?

JoeRobich: true. we just tell azdo to retain our signed builds.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -88,8 +88,8 @@ private static async Task<Build> GetLatestBuildAsync(CancellationToken cancellat
             var builds = await GetBuildsFromTFSAsync(buildClient, definitions, cancellationToken, resultFilter: null);
 
             return (await GetInsertableBuildsAsync(buildClient, cancellationToken, from build in builds
-                    orderby build.FinishTime descending
-                    select build)).FirstOrDefault();
+                                                                                   orderby build.FinishTime descending
```
RikkiGibson: ehhh I understand if you don't bother here but it feels like the arguments to GetInsertableBuildsAsync should be wrapped here.

JoeRobich: gonna have to change the way this is wrapped because I don't want to revert the outdent only to hit it again with future formats.

PR #724, opened 2020-02-24
Use PR description for merge commit message

### LINKED ISSUE #708: Make the RIT merge commit more descriptive
This is a standard message for every insertion this tool is making now. Instead it could include PR # and/or title
https://github.com/dotnet/roslyn-tools/blob/c8fc9dd217ce2fb69e2ac3bca70409300a0c8db1/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs#L237

JoeRobich:
Instead of "RIT Auto-Complete" use messages like "Updating Roslyn from 20200224.1 (88dd6e5) to 20200224.2 (d51adc7)"

Fixes #708

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
@@ -282,9 +282,7 @@ namespace Roslyn.Insertion
 
                 // ********************* Create pull request *****************************
                 var oldBuild = await GetSpecificBuildAsync(oldComponentVersion, cancellationToken);
-
                 var prDescriptionMarkdown = CreatePullRequestDescription(oldBuild, buildToInsert, useMarkdown: true);
-                var prDescriptionText = CreatePullRequestDescription(oldBuild, buildToInsert, useMarkdown: false);
 
                 if (buildToInsert.Result == BuildResult.PartiallySucceeded)
                 {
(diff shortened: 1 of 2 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -388,7 +386,7 @@ public static partial class RoslynInsertionTool
                     Console.WriteLine($"Set PR to Auto-Complete");
                     try
                     {
-                        await SetAutoCompleteAsync(pullRequest, cancellationToken);
+                        await SetAutoCompleteAsync(pullRequest, prDescriptionText, cancellationToken);
```
RikkiGibson: consider declaring this closer to the usage.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -473,7 +471,24 @@ public static GitChange GetChangeOpt(string path, string originalText, string ne
             }
         }
 
-        public static string GetBuildDescription(this Build build)
+        private static string CreatePullRequestDescription(Build oldBuild, Build newBuild, bool useMarkdown)
```
dmonroym: >newBuild [](start = 81, length = 8)

Can we either rename newBuild to buildToInsert or the other way around? They should be referring to the same build always so we should keep the naming consistent

dmonroym:
:shipit:

Issue #364, opened 2018-10-05
Make separate schedules for GitHubCreateMergePRs based on frequency


dpoeschl:
Some merges should be happening basically all the time (`*` --> `*-vs-deps`). Others should be more spaced out (most `feature/*` branches).

RikkiGibson:
Do we still want this? Seems like opinions differ about how often we should merge to -vs-deps.

PR #728, opened 2020-02-25
Add daily merge frequency to GHCreateMergePR

JoeRobich:
Fixes https://github.com/dotnet/roslyn-tools/issues/364

### DIFF
--- a/src/GitHubCreateMergePRs/Program.cs
+++ b/src/GitHubCreateMergePRs/Program.cs
@@ -186,12 +186,20 @@ public class Program
         // Since this is run on AzDO as an automated cron pipeline, times are in UTC.
         // See https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#scheduled-triggers
         var runDateTime = DateTime.UtcNow;
-        var currentDate = runDateTime.Date;
 
         // Since cron should schedule this to run every 3 hours starting at 12am,
-        // we can check if we are running within the first 15 minutes of the day.
-        var fifteenMinutes = new TimeSpan(hours: 0, minutes: 15, seconds: 0);
-        var isStartOfDay = runDateTime - currentDate < fifteenMinutes;
+        // we expect to be run within a 10 minute window of this time.
+
+        // Adjust the time 5 minutes into the future in case the pipeline machine
+        // and scheduler machine have mismatched clocks.
+        var adjustedRunDateTime = runDateTime.AddMinutes(5);
+        var adjustedRunDate = adjustedRunDateTime.Date;
+
+        // Because of the adjusted run time is being used we treat a run as valid
+        // if it begins in the last 5 minutes of the previous day through the first
+        // 5 minutes of the current day.
+        var tenMinutes = new TimeSpan(hours: 0, minutes: 10, seconds: 0);
+        var isStartOfDay = adjustedRunDateTime - adjustedRunDate < tenMinutes;
 
         // Daily and Weekly runs only happen at the start of the day
         if (!isStartOfDay)
@@ -199,7 +207,7 @@ public class Program
             return false;
         }
 
-        var isSunday = currentDate.DayOfWeek != DayOfWeek.Sunday;
+        var isSunday = adjustedRunDate.DayOfWeek != DayOfWeek.Sunday;
 
         // Weekly runs only happen on Sunday
         if (frequency == "weekly")
--- a/src/GitHubCreateMergePRs/Program.cs
+++ b/src/GitHubCreateMergePRs/Program.cs
@@ -207,12 +207,10 @@ public class Program
             return false;
         }
 
-        var isSunday = adjustedRunDate.DayOfWeek != DayOfWeek.Sunday;
-
         // Weekly runs only happen on Sunday
         if (frequency == "weekly")
         {
-            return isSunday;
+            return adjustedRunDate.DayOfWeek == DayOfWeek.Sunday;
         }
 
         // Daily runs happen every day
--- a/src/GitHubCreateMergePRs/Program.cs
+++ b/src/GitHubCreateMergePRs/Program.cs
@@ -115,6 +115,10 @@ public class Program
 
     private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool isDryRun, string githubToken)
     {
+        // Since this is run on AzDO as an automated cron pipeline, times are in UTC.
+        // See https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#scheduled-triggers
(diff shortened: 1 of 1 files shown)


JoeRobich:
@RikkiGibson @dpoeschl PTAL

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -157,4 +156,53 @@ private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool i
...
+        var frequency = merge.Attribute("frequency")?.Value.ToLower();
+
+        // We always run when a frequency isn't specified
+        if (string.IsNullOrEmpty(frequency))
+        {
+            return true;
+        }
+
+        // We always run when an unexpected frequency value is specified
+        if (frequency != "daily" && frequency != "weekly")
```
dibarbet: should we run, or not run at all and just error out the job (so we realize there is an issue)?

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -157,4 +156,53 @@ private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool i
             }
         }
     }
+
+    private static bool ShouldRunMerge(XElement merge, bool isAutomatedRun)
```
dibarbet: should we doc what frequencies are supported?
as far as I can tell it's daily and weekly, but perhaps we should make it obvious.

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -157,4 +156,53 @@ private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool i
...
+
+        // We always run when an unexpected frequency value is specified
+        if (frequency != "daily" && frequency != "weekly")
+        {
+            Console.WriteLine($"##vso[task.logissue type=error]Unexpected merge frequency specified ({frequency}).");
+
+            return true;
+        }
+
+        var runDateTime = DateTime.Now;
```
dibarbet: Should we make sure to get this in UTC always (in case the machine time changes)?  the pipeline cron always uses UTC - https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#scheduled-triggers

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -157,4 +156,53 @@ private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool i
...
+        var currentDate = runDateTime.Date;
+
+        // Are we within the first 15 minutes of the day?
+        var fifteenMinutes = new TimeSpan(hours: 0, minutes: 15, seconds: 0);
```
dibarbet: Might be good to make a note that we can do this check b/c the current pipeline cron is running every 3 hours starting at 12am (so it will run at 12am every day).

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -157,4 +156,53 @@ private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool i
...
+        var isStartOfDay = runDateTime - currentDate < fifteenMinutes;
```
dibarbet: If the machine time is running fast, should we allow 15min leeway into the previous day as well?

src/GitHubCreateMergePRs/Program.cs
dibarbet: if we error the job it *should* be caught in a PR check I think which would be good.

src/GitHubCreateMergePRs/Program.cs
JoeRobich: That seems excessive. =p

JoeRobich: Although I could take the `runDateTime` and `.AddMinutes(5)` then do all my calculations based on the adjusted DateTime to keep things simple...

JoeRobich: But really I just hope that triggers don't fire before the trigger time...

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -157,4 +156,66 @@ private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool i
...
+        var tenMinutes = new TimeSpan(hours: 0, minutes: 10, seconds: 0);
+        var isStartOfDay = adjustedRunDateTime - adjustedRunDate < tenMinutes;
+
+        // Daily and Weekly runs only happen at the start of the day
+        if (!isStartOfDay)
+        {
+            return false;
+        }
+
+        var isSunday = adjustedRunDate.DayOfWeek != DayOfWeek.Sunday;
```
RikkiGibson: Please do this inside the body of the `if (frequency == "weekly")`.

Also, it looks like this condition should maybe be negated.

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -157,4 +156,64 @@ private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool i
...
+
+        // Throw when an unexpected frequency value is specified
+        if (frequency != "daily" && frequency != "weekly")
+        {
+            throw new Exception($"Unexpected merge frequency specified: '{frequency}'. Valid values are 'daily' and 'weekly'.");
+        }
+
+        // Since this is run on AzDO as an automated cron pipeline, times are in UTC.
+        // See https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#scheduled-triggers
+        var runDateTime = DateTime.UtcNow;
```
RikkiGibson: Please get DateTime.UtcNow only once in a run of the tool. e.g. by passing in a `DateTime currentTime` parameter.

JoeRobich: Doh! good catch

src/GitHubCreateMergePRs/Program.cs
```diff
@@ -157,4 +160,60 @@ private static async Task RunAsync(XDocument config, bool isAutomatedRun, bool i
...
+
+        // Since cron should schedule this to run every 3 hours starting at 12am,
+        // we expect to be run within a 10 minute window of this time.
+
+        // Adjust the time 5 minutes into the future in case the pipeline machine
+        // and scheduler machine have mismatched clocks.
+        var adjustedRunDateTime = runDateTime.AddMinutes(5);
+        var adjustedRunDate = adjustedRunDateTime.Date;
+
+        // Because the adjusted run time is being used, we treat a run as valid if
```
RikkiGibson: was this heuristic decided on by looking at the actual start times for the pipeline?

JoeRobich: Paranoia. The pipeline seems to be very consistent from a cursory look at start times.

PR #743, opened 2020-03-09
Add link to PR for validation build

genlu:
The URL will be displayed in insertion description if build has Tag `PRNumber:[PR number]`


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -504,7 +505,31 @@ private static string CreatePullRequestDescription(Build oldBuild, Build buildTo
...
+
+        private static string GetGitHubPullRequestUrlMessage(Build build, bool useMarkdown)
+        {
+            var prValidationMessage = string.Empty;
+
+            if (build.Repository.Type == "GitHub")
+            {
+                var repoURL = $"http://github.com/{build.Repository.Id}";
+
+                var tagPrefix = "PRNumber:";
```
RikkiGibson: microscopic nit: consider using 'const string' here

PR #770, opened 2020-04-24
Update insertion PR title when overwriting

JoeRobich:
Who hasn't forgotten to update an overwritten dummy PR's title before inserting?


JoeRobich:
Before:
image

After Overwrite:
image

RikkiGibson:
only concern I have is maybe the dummy PR gets lost in the noise after overwriting. but it is likely to be the only PR to some QB branch.

JoeRobich:
> maybe the dummy PR gets lost in the noise after overwriting

It'll definitely still have its QB approval to help it stand out. Plus if it were still a concern, a prefix could be entered when overwriting.

PR #772, opened 2020-05-01
Update GitHubCreateMergePRs/config.xml for recent language feature branch changes

gafter:
- Delete features/patterns3 merge
- Add merge for features/target-typing
- Add merge for features/covariant-returns

### DIFF
--- a/src/GitHubCreateMergePRs/config.xml
+++ b/src/GitHubCreateMergePRs/config.xml
@@ -28,7 +28,8 @@
 
     <!-- Roslyn feature branches -->
     <merge from="master" to="features/static-lambdas" owners="jcouv" />
-    <merge from="master" to="features/patterns3" owners="gafter" frequency="weekly" />
+    <merge from="master" to="features/covariant-returns" owners="gafter" />
+    <merge from="master" to="features/target-typing" owners="gafter" />
     <merge from="master" to="features/source-generators" owners="chsienki" frequency="weekly" />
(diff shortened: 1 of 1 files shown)


gafter:
@agocke @JoeRobich @dibarbet Could one of you please look at this to make sure I'm not doing something terribly stupid?

src/GitHubCreateMergePRs/config.xml
```diff
@@ -28,9 +28,11 @@
 
     <!-- Roslyn feature branches -->
     <merge from="master" to="features/static-lambdas" owners="jcouv" />
-    <merge from="master" to="features/patterns3" owners="gafter" frequency="weekly" />
+    <merge from="master" to="features/covariant-returns" owners="gafter" />
+    <merge from="master" to="features/target-typing" owners="gafter" />
```
RikkiGibson, line 32: This may result in merges being attempted once every 3 hours. If this is too often to be manageable for you, you may want to add `frequency="daily"` or `frequency="weekly"`.

gafter: I'm not worried about the frequency.  It has the advantage that if one blocks, a later one merging will close all the earlier ones. Which will make it easier to keep up to date without much work.

333fred: Confirmed with @gafter that this was fine.

PR #778, opened 2020-05-23
Write a tool to tag commits that shipped in VS

jasonmalinowski:
This tool walks the internal VS repo and sees what builds are tagged there, and then tags the matching builds in our repo.


jasonmalinowski:
@JoeRobich has already used this tool once successfully, so we'll call that good for testing.

RikkiGibson:
Is there any release pipeline that will run this tool automatically on the server side?

jasonmalinowski:
@RikkiGibson: no, and it creates the tags locally as that was how I was validating it's correctness prior to pushing.

PR #783, opened 2020-05-28
Fix the checks for whether existing PR has merge conflicts

JoeRobich:


JoeRobich:
> make sure any other usages of IsPrConflicted are correct 

We are looking good.

> Have you manually tested this change?

Took a bit of work to test but should be easier next time.

PR #784, opened 2020-06-03
Run roslyn merge tool every 5 hours instead

genlu:
Given reduced capability in our machine pool, also reduce frequency to allow CI to have a better change to finish before being reset.

@JoeRobich @RikkiGibson


genlu:
This doesn't seem to be a big problem, closed.

PR #789, opened 2020-06-13
Use more of the available characters in the message limit

RikkiGibson:
The last insertion used 3581 of 4000 available characters. It feels like we could fit more information in the description.

BTW, this `http://github.com` -> `//github.com` change does not affect the way the link gets copied when you right click in the browser. You still get an https prefix so it shouldn't introduce problems in practice.

Baseline insertion: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/254613
Insertion with this change: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/254665

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -560,7 +560,7 @@ namespace Roslyn.Insertion
                     .Reverse()
                     .ToList();
 
-                return (result, $"https://github.com/{repoId}/compare/{fromSHA}...{toSHA}?w=1");
+                return (result, $"//github.com/{repoId}/compare/{fromSHA}...{toSHA}?w=1");
             }
 
             throw new NotSupportedException("Only builds created from GitHub repos support enumerating commits.");
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -579,7 +579,7 @@ namespace Roslyn.Insertion
 
             var description = new StringBuilder(prDescription + Environment.NewLine);
 
-            var repoURL = $"http://github.com/{oldBuild.Repository.Id}";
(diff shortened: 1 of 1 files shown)


RikkiGibson:
PTAL @JoeRobich @dibarbet

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -675,16 +677,29 @@ internal static string AppendChangesToDescription(string prDescription, Build ol
...
                 }
+                else
+                {
+                    description.AppendLine(prLink);
+                }
             }
 
-            return description.ToString();
+            var result = description.ToString();
+            if (result.Length > hardLimit)
```
JoeRobich, line 696: Will this ever be true if we are breaking before we hit the hard limit?

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -675,16 +677,29 @@ internal static string AppendChangesToDescription(string prDescription, Build ol
                     prLink = $@"- [{comment}]({repoURL}/commit/{fullSHA})";
                 }
 
-                description.AppendLine(prLink);
+                const string limitMessage = "Changelog truncated due to description length limit.";
```
JoeRobich, line 680: Should we make this less verbose? "Changelog too long" or "... =("

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
RikkiGibson: I don't expect to really hit this. I added it as more of a guardrail, so if the insertion pipeline ever fails, we might be able to diagnose it more easily

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
RikkiGibson: My vote is for "... you haven't inserted in a while, huh?"

PR #791, opened 2020-06-29
Update code flow for 16.7 p4 snap

dibarbet:


src/GitHubCreateMergePRs/config.xml
```diff
@@ -11,25 +11,22 @@
     <merge from="release/dev16.6" to="master" />
```
JoeRobich, line 11: I think at this point we can flow 16.6 into 16.7. We were only trying to avoid merging into 16.7-p2

PR #792, opened 2020-06-30
Fix description length check

dibarbet:
Currently failing
https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-environment-logs&releaseId=727349&environmentId=3972420


RikkiGibson:
Oh wow, I wouldn't have thought of that :eyes:

I also may PR the change from https://github.com/dotnet/roslyn-tools/compare/master...RikkiGibson:truncate-description once I verify it fixes the problem. Basically if anything ever causes the description to be too long we should warn, truncate, and carry on.

PR #795, opened 2020-07-21
Add arcade publishing step

JoeRobich:
Test build showing publishing stages will run https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3917202&view=results


JoeRobich:
> approved assuming the test build completes and publishes correctly

Since I didn't register the pr branch to a channel it won't publish. But this certainly adds the publishing step to our build. =)

PR #798, opened 2020-07-29
Stop publishing artifacts to drop path

JoeRobich:
<strike>Borrowed the build yml from sourcelink (https://github.com/dotnet/sourcelink/blob/master/azure-pipelines.yml)</strike>

I stopped publishing to the drop path since that was causing the build to fail.


JoeRobich:
Merging to get builds passing

PR #801, opened 2020-08-01
Add /cherrypick option

### LINKED ISSUE #800: Include commits to cherry-pick in an insertion
The cherry-pick UI in AzDo is a pain, and so is pulling down VS branches locally in order to cherry-pick commits across them. We should update the insertion tool to accept a list of commits which should be cherry-picked into the insertion branch.

This could be a queue-time variable with a default value, since we have a different release definition for each Roslyn branch that we want to automatically create insertions for. This will make life easier if we're in a situation where many builds need the same commits cherry-picked in. For example, we may have a Roslyn change which requires a VS-side change, but we also need to fix an RPS regression, and thus need to make reverts or minor edits to the compiler.

/cc @333fred.

RikkiGibson:
Closes #800

See https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/264858?_a=commits for an example insertion

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool.Commandline/RoslynInsertionToolCommandline.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool.Commandline/RoslynInsertionToolCommandline.cs
@@ -221,7 +221,7 @@ partial class RoslynInsertionToolCommandline
             },
             {
                 "cherrypick=",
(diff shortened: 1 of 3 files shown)


RikkiGibson:
I need this feature so I am merging. Feel free to add any feedback and we can address at a later date.

PR #804, opened 2020-08-10
Include nuget package ID in tag

davidwengier:
For VS for Mac insertions being able to the see the commit hash and what version of VS things went in to is great, but I need to reference the nuget package, so adding that info in would be helpful. I think.

I'm not entirely sure though, as I also need the info for VS master which _isn't_ part of this tagging..


davidwengier:
Thinking about this more, without master this tool is of limited use.. I might just borrow the code a bit instead :)

PR #809, opened 2020-08-14
Add some helpful inline documentation

RikkiGibson:
/cc @JoeRobich 

The goal of this change is to make the preferred first-resort workflow easier to get going with :smile:


RikkiGibson:
I think we're going to try and change this to an aka.ms link and maybe condition it on Roslyn only.

RikkiGibson:
@JoeRobich what do you think?

RikkiGibson:
https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/271059

PR #820, opened 2020-09-04
Handle spaces in /cherrypick argument

RikkiGibson:
/cc @allisonchou 

Will try making a test insertion with this change if I get the chance.


allisonchou (approved):

RikkiGibson:
I think this is a low risk change so will just merge. If things go bad the next time we try to use /cherrypick I think we'll know where to look :)

PR #827, opened 2020-09-18
Add tool to generate table of hashes for CodeActions

JoeRobich:
This tool generates a table of ActionName, (Hash) Prefix, and (Hash) Suffix which can be used from a Kusto query to decode lightbulb session telemetry events. When no commandline arguments are passed, this tool generates the hashes for the Roslyn code actions. When passing arguments, they should be space separated paths to the assemblies whose code actions should be included in the table.


JoeRobich:
CC: @ryzngard, @mikadumont

JoeRobich:
Going to move this tool to the Roslyn repo.

JoeRobich:
See https://github.com/dotnet/roslyn/pull/47855

PR #849, opened 2020-11-05
Remove RepoToolset

### LINKED ISSUE #793: Migrate packages to ADO from roslyn-tools feeds
We publish and consume packages from the roslyn-tools myget feed that aren't used by Roslyn.  These need to be moved as well once we finish moving Roslyn.

RikkiGibson:
Closes #793

Removes the RepoToolset project which has spun off into dotnet/arcade. This contains the last mentions of myget in this repo.

PTAL @tmat


dmonroym:
Yes, project system still uses this and we have no plans of moving onto arcade. Can the feeds be changed instead of deleting all this?

RikkiGibson:
Yep, thanks for the heads up :)

PR #898, opened 2021-02-17
Add list patterns merges

333fred:
Adds list pattern merges. @alrz is a non-MSFT contributor but will be doing the implementation here, so I'm adding him to be notified when the merge bot encounters errors. alrz, you won't be able to resolve these errors on your own, but I can work with you to make sure merges end up correct.


333fred:
@RikkiGibson @jaredpar for review

333fred:
After looking at the code, we can't do this because alrz can't be assigned to things. @alrz, when we need to do a merge I'll tag you manually.

RikkiGibson (approved):

alrz:
(This might help in the future, just in case https://github.blog/changelog/2019-06-25-assign-issues-to-issue-commenters/)

PR #904, opened 2021-03-06
Read component version from local manifests when possible

JoeRobich:
As part of moving Roslyn's build, we will not have access to domain resources. We have been pulling copies of the component manifest when determining component version, but those same manifests are available from our insertion artifacts.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/InsertionArtifacts.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/InsertionArtifacts.cs
@@ -1,12 +1,22 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
 
+using System.IO;
+using System.Linq;
+
 namespace Roslyn.Insertion
 {
     internal abstract class InsertionArtifacts
     {
+        internal abstract string RootDirectory { get; }
+
         public abstract string GetPackagesDirectory();
         public abstract string GetDependentAssemblyVersionsFile();
         public abstract string[] GetOptProfPropertyFiles();
+
+        public string FindFilePath(string fileName)
+        {
+            return Directory.EnumerateFiles(RootDirectory, fileName, SearchOption.AllDirectories).SingleOrDefault();
+        }
     }
 }
--- a/src/RoslynInsertionTool/RoslynInsertionTool/InsertionArtifacts.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/InsertionArtifacts.cs
@@ -1,22 +1,13 @@
 // Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
 
-
-using System.IO;
-using System.Linq;
-
 namespace Roslyn.Insertion
 {
     internal abstract class InsertionArtifacts
     {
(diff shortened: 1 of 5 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/LegacyInsertionArtifacts.cs
```diff
@@ -32,14 +32,14 @@ public static bool TryCreateFromLocalBuild(string buildDirectory, out InsertionA
         public override string GetPackagesDirectory()
         {
             // For example: "\\cpvsbuild\drops\Roslyn\Roslyn-Main-Signed-Release\20160315.3\DevDivPackages"
```
dibarbet, line 34: This comment should not apply anymore right, since we can't read from here

src/RoslynInsertionTool/RoslynInsertionTool/InsertionArtifacts.cs
```diff
@@ -1,12 +1,22 @@
...
 {
     internal abstract class InsertionArtifacts
     {
+        internal abstract string RootDirectory { get; }
+
         public abstract string GetPackagesDirectory();
         public abstract string GetDependentAssemblyVersionsFile();
         public abstract string[] GetOptProfPropertyFiles();
+
+        public string FindFilePath(string fileName)
```
RikkiGibson, line 17: I feel like it could be more clear to inline this at the point of usage. YMMV.

RikkiGibson (approved):

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -423,31 +423,47 @@ private static async Task<Component[]> GetComponents(string[] urls)
 
                 var fileName = urlString.Split(';').Last();
                 var name = fileName.Remove(fileName.Length - 6, 6);
-                var version = await GetVersionFromComponentUrl(uri);
+                var localFilePath = buildArtifacts.FindFilePath(fileName);
```
dibarbet, line 426: do you have an example of finding the path?  I'm not 100% certain why we can't 'build' the path if that makes sense

dibarbet (approved):

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
JoeRobich: The vsman files do not seem to be in a consistent location other than in the VSSetup artifact. Sometimes you will find them in the DevDivPackages subfolder, sometimes in an Insertions subfolder.

src/RoslynInsertionTool/RoslynInsertionTool/LegacyInsertionArtifacts.cs
JoeRobich: I'm not sure when this comment would have applied. Even when working with legacy artifacts today they are either downloaded as part of the pipeline trigger or we download them from the build. We don't try to access them from a share.

src/RoslynInsertionTool/RoslynInsertionTool/InsertionArtifacts.cs
JoeRobich: Sure.

Issue #854, opened 2020-11-16
Consolidate auto-insertion pipelines


RikkiGibson:
There's a significant amount of "sprawl" across our insertion pipelines--one for each release branch, one for PR validation, one for master validation, etc.

It would be great if we could consolidate all the pipelines that create draft insertions, for example. The insertion tool should be able to handle the variations between e.g. official builds for QB branches and validation builds by just looking at the tags on the build.

Ideally we could have a pipeline that creates draft/validation insertions and another pipeline that creates auto-merging insertions.

RikkiGibson:
This is somewhat addressed by the Roslyn changes to perform insertions as part of the official build. Keeping open in case we decide to do something similar with OneOffInsertion/UpdateExistingPr/etc.

PR #906, opened 2021-03-08
Move RIT to build against net472 and net6.

JoeRobich:
This change would set us up to move to a dotnet-tool. I tried to target netstandard1.3 but our dependencies didn't want to cooperate. 

Moving from net46 to net472 does necessitate updating insertion pipelines to invoke RIT from the correct folder. This has been completed for Roslyn and Roslyn-SDK. Would need to coordinate with the project-system & the testing teams to make sure no one gets broken.


src/RoslynInsertionTool/RoslynInsertionTool.Commandline/RIT.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <RootNamespace>RoslynInsertionTool</RootNamespace>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net46</TargetFramework>
+    <TargetFrameworks>net472;net5</TargetFrameworks>
```
RikkiGibson, line 6: wait, `net5` is a TFM? not `net5.0`?

RikkiGibson (approved):
Looks fine to me, however, wondering if we can drop the netframework target entirely. Does that make deployment harder than it currently is?

JoeRobich:
> Looks fine to me, however, wondering if we can drop the netframework target entirely. Does that make deployment harder than it currently is?

It is a bigger change, but not necessarily a hard change. Still need to check with @dmonroym and @shyamnamboodiripad that the current change will be low impact enough to get completed.

shyamnamboodiripad:
Tagging @AbhitejJohn as well. @JoeRobich I don't think it would be too difficult to modify the pipelines for UT insertion as they are based on corresponding roslyn pipelines - but Abhitej should have more context on this.

We may also need to update the oneoffinsertion, updatexistingpr, overwriteexistinpr queues. Joey - will you be fixing those?

JoeRobich:
> We may also need to update the oneoffinsertion, updatexistingpr, overwriteexistinpr queues. Joey - will you be fixing those?

I've already updated those pipelines to handle this change. Basically I update the powershell script to test for a net472 directory before falling back to the current net46.

```powershell
mv RoslynTools.VisualStudioInsertionTool.* RIT

if (Test-Path .\RIT\tools\net472) {
  cd .\RIT\tools\net472
}
else {
  cd .\RIT\tools\net46
}

.\OneOffInsertion.ps1 ...
```

RikkiGibson:
Should we attempt to do #854 before merging this change?

RikkiGibson:
Basically, there is so much copy-pasting in web UIs to configure our pipelines. If we just had a YAML template for "obtain RIT" and "run RIT" tasks, this kind of maintenance could potentially get easier. I don't know for sure if AzDO supports our particular use case with this, though.

JoeRobich:
> Should we attempt to do #854 before merging this change?

I've updated all of Roslyn's insertion pipelines.

JoeRobich:
Sigh, I goofed when I updated Roslyns insertion pipelines. I shouldn't have changed the working directory. Updated again with the intention of removing the `if` when this merges.
```powershell
mv RoslynTools.VisualStudioInsertionTool.* RIT
if (Test-Path .\RIT\tools\net472) {
  .\RIT\tools\net472\OneOffInsertion.ps1 ...
}
else {
  .\RIT\tools\net46\OneOffInsertion.ps1 ...
}
```

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.csproj
```diff
@@ -1,35 +1,16 @@
...
-    <PackageReference Include="Microsoft.VisualStudio.Services.Client" Version="16.155.0-preview" />
-    <PackageReference Include="Microsoft.VisualStudio.Services.InteractiveClient" Version="16.155.0-preview" />
-    <PackageReference Include="Microsoft.VisualStudio.Services.Release.Client" Version="16.155.0-preview" />
-    <PackageReference Include="Microsoft.WindowsAzure.ConfigurationManager" Version="3.2.3" />
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
-    <PackageReference Include="NuGet.Versioning" Version="4.9.4" />
-    <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="5.5.0" />
-    <PackageReference Include="System.IO.Compression.ZipFile" Version="4.3.0" />
-    <PackageReference Include="System.ValueTuple" Version="4.4.0" />
-    <PackageReference Include="WindowsAzure.ServiceBus" Version="5.2.0" />
```
JoeRobich, line 30: Thanks Remove Unused Reference feature!

RikkiGibson: Yeah whoever added that is great

src/RoslynInsertionTool/RoslynInsertionTool.Commandline/RIT.csproj
JoeRobich: Updated to `net5.0`

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.csproj
JoeRobich: Thanks @AlekseyTs!

RikkiGibson:
Hi Joey, how is it going with this PR? Please let me know if there's any way I can assist with it.

JoeRobich:
> Hi Joey, how is it going with this PR?

This PR should be ready to merge. I'm just hesitant to break everyone. =p

dmonroym:
> > Hi Joey, how is it going with this PR?
> 
> This PR should be ready to merge. I'm just hesitant to break everyone. =p

@RikkiGibson perhaps we can publish the package with this update elsewhere and dogfood it for a bit before publishing it in dotnet-eng

RikkiGibson:
Sure, it feels like we could publish to some testing feed and use the "OneOffInsertion - Experimental" pipeline to try out this change.

RikkiGibson:
I think most people run the tool via the powershell scripts, so I'm optimistic things won't be broken too badly by this change. Maybe we can merge in a week or two?

global.json
```diff
@@ -1,13 +1,8 @@
 {
   "tools": {
-    "dotnet": "6.0.100",
-    "runtimes": {
-      "dotnet": [
-        "3.1.20"
-      ]
-    }
+    "dotnet": "6.0.100"
```
JoeRobich, line 3:
```suggestion
    "dotnet": "6.0.100",
    "runtimes": {
      "dotnet": [
        "3.1.20"
      ]
    }
```

global.json
```diff
@@ -10,4 +10,4 @@
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "7.0.0-beta.21573.3"
   }
-}
+}
```
JoeRobich, line 13:
```suggestion
}

```

JoeRobich:
@RikkiGibson Think we should move forward with this? One thought is we could rework this tool with the System.CommandLine api and convert the ps1 scripts to be subcommands of the RIT tool.

RikkiGibson:
Yeah, let's do this. It seems like this PR is a first step toward what you've described.

JoeRobich:
Instead of taking this PR (https://github.com/dotnet/roslyn-tools/pull/906) we should move RIT functionality to dotnet-roslyn (https://github.com/dotnet/roslyn-tools/issues/1168) as a set of subcommands which correspond to various scripts `create-insertion` (OneOffInsertion), `create-dummy-insertion` (CreateDummyPR), `update-insertion`, & `overwrite-insertion` (UpdateExistingPR).

Closing this PR.

RikkiGibson:
Should we just copy-paste the parts of the implementation we need into a new command under the dotnet-roslyn CLI tool here? I recall having problems trying to share code between this and a dotnet CLI tool project.

JoeRobich:
We shouldn't try to share code. I would argue that we remove the old RIT implementation when the code is moved. Users of RIT won't be broken by the change but would then need to move to get new functionality.

RikkiGibson:
I think I'm fine with removing. It shouldn't be hard for users to move from the old to new tool if they need some fix.

This might be a good opportunity for us to start using semantic versioning of the CLI tool, i.e. bumping the major version when we make breaking changes.

PR #935, opened 2021-04-09
Resume dev17 codeflow

genlu:
Now release/dev17.0 is targeting preview 2, we should resume the codeflow from main.

@sharwell @JoeRobich @dibarbet

### DIFF
--- a/src/GitHubCreateMergePRs/config.xml
+++ b/src/GitHubCreateMergePRs/config.xml
@@ -11,11 +11,11 @@
 
     <!-- Roslyn non-vs-deps branches (flowing between releases) -->
     <merge from="release/dev16.10" to="main" />
-    <!--merge from="main" to="release/dev17.0" /-->
+    <merge from="main" to="release/dev17.0" />
 
     <!-- Roslyn vs-deps branches (flowing between releases) -->
     <merge from="release/dev16.10-vs-deps" to="main-vs-deps" />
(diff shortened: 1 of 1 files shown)


ryzngard (approved):

src/GitHubCreateMergePRs/config.xml
```diff
@@ -11,11 +11,12 @@
...
     <!-- Roslyn non-vs-deps branches (flowing between releases) -->
     <merge from="release/dev16.10" to="main" />
-    <!--merge from="main" to="release/dev17.0" /-->
+    <merge from="main" to="release/dev17.0" />
 
     <!-- Roslyn vs-deps branches (flowing between releases) -->
     <merge from="release/dev16.10-vs-deps" to="main-vs-deps" />
-    <!--merge from="main-vs-deps" to="release/dev17.0-vs-deps" /-->
+    <merge from="main-vs-deps" to="release/dev17.0-vs-deps" />
+    <merge from="release/dev17.0-preview1-vs-deps" to="release/dev17.0-vs-deps" />
```
dibarbet, line 19: Do we want this if we're not merging main into preview1?  Can't tell if that might just break things more

genlu: We are not inserting preview 2 anywhere, so I think this is fine. Get the code flowing from p1 to p2 would ensure p2 has all the fixes included in p1 branch.

dibarbet (approved):

PR #943, opened 2021-04-17
Pass ComponentUserName when ComponentAzdoUri specified

JoeRobich:
Missed a parameter in https://github.com/dotnet/roslyn-tools/pull/942

### DIFF
--- a/src/RoslynInsertionTool/scripts/HelperFunctions.ps1
+++ b/src/RoslynInsertionTool/scripts/HelperFunctions.ps1
@@ -29,6 +29,14 @@ function GetComponentProjectName([string] $componentProjectName) {
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/RoslynInsertionTool/scripts/HelperFunctions.ps1
```diff
@@ -29,6 +29,14 @@ function GetComponentProjectName([string] $componentProjectName) {
     }
 }
 
+function GetComponentUserName([string] $componentAzdoUri) {
+    if (IsDefaultValue $componentAzdoUri) {
+        return ""
+    } else {
+        return "/cbu=vslsnap@microsoft.com"
```
JoeRobich, line 36: The scripts pass the user hardcoded already but the component user needs to be optionally included.

PR #952, opened 2021-04-26
Update flow for d17

dibarbet:


src/GitHubCreateMergePRs/config.xml
```diff
@@ -11,17 +11,14 @@
 
     <!-- Roslyn non-vs-deps branches (flowing between releases) -->
     <merge from="release/dev16.10" to="main" />
-    <merge from="main" to="release/dev17.0" />
```
dibarbet, line 14: I removed the flow to the dev17 branch.  Once we snap for a p1 we can add flow back (from d16 -> d17 -> main)

src/GitHubCreateMergePRs/config.xml
```diff
@@ -11,17 +11,14 @@
...
 
     <!-- Roslyn vs-deps branches (flowing between releases) -->
     <merge from="release/dev16.10-vs-deps" to="main-vs-deps" />
-    <merge from="main-vs-deps" to="release/dev17.0-vs-deps" />
-    <merge from="release/dev17.0-preview1-vs-deps" to="release/dev17.0-vs-deps" />      
+    <merge from="release/dev17.0-preview1-vs-deps" to="main-vs-deps" />
```
dibarbet, line 17: theoretically we should be flowing from release/dev16.10-vs-deps -> release/dev17.0-preview1-vs-deps -> main-vs-deps, but we're not ready to take changes in release/dev17.0-preview1-vs-deps, right @genlu ?

allisonchou (approved):

src/GitHubCreateMergePRs/config.xml
genlu: Yes, we'd like to wait a couple more days until the interop work is considered done, should be able to resume by the end of this week. So maybe we can start creating the merge PR and fixing potential conflicts before then?

PR #984, opened 2021-05-26
Stop 16.11 to 17.0 P1

Cosifne:
I just have an offline discussion with @vatsalyaagrawal and we want to pause the change from 16.11 to 17.0 P1

### DIFF
--- a/src/GitHubCreateMergePRs/config.xml
+++ b/src/GitHubCreateMergePRs/config.xml
@@ -15,6 +15,7 @@
 
     <!-- Roslyn vs-deps branches (flowing between releases) -->
     <merge from="release/dev16.10-vs-deps" to="release/dev16.11-vs-deps" />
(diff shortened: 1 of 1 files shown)


src/GitHubCreateMergePRs/config.xml
```diff
@@ -15,7 +15,6 @@
 
     <!-- Roslyn vs-deps branches (flowing between releases) -->
     <merge from="release/dev16.10-vs-deps" to="release/dev16.11-vs-deps" />
-    <merge from="release/dev16.11-vs-deps" to="release/dev17.0-preview1-vs-deps" />
```
dibarbet, line 18: dev16.11-vs-deps changes won't flow into `main-vs-deps` - should this just be changed from `release/dev17.0-preview1-vs-deps` to `main-vs-deps`

dibarbet (approved):
Nevermind, beat me to it :)

PR #991, opened 2021-06-03
Add SkipCoreXTPackages option

jakubch1:
This is adding possiblity to skip some packages from upgrading in corext.
Before this change there was hardcoded package VS.Tools.Roslyn which was not updating.
This change is adding setting SkipCoreXTPackages which is the list of packages to be skipped. By default it contains VS.Tools.Roslyn.

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool.Commandline/Settings.settings
+++ b/src/RoslynInsertionTool/RoslynInsertionTool.Commandline/Settings.settings
@@ -71,5 +71,8 @@
     <Setting Name="VisualStudioRepoAzdoUri" Type="System.String" Scope="Application">
       <Value Profile="(Default)">https://devdiv.visualstudio.com/DefaultCollection</Value>
     </Setting>
+    <Setting Name="SkipCoreXTPackages" Type="System.String" Scope="Application">
+      <Value Profile="(Default)">VS.Tools.Roslyn</Value>
+    </Setting>
   </Settings>
 </SettingsFile>
--- a/src/RoslynInsertionTool/RoslynInsertionTool.Commandline/Settings.settings
+++ b/src/RoslynInsertionTool/RoslynInsertionTool.Commandline/Settings.settings
@@ -72,7 +72,7 @@
       <Value Profile="(Default)">https://devdiv.visualstudio.com/DefaultCollection</Value>
     </Setting>
     <Setting Name="SkipCoreXTPackages" Type="System.String" Scope="Application">
-      <Value Profile="(Default)">VS.Tools.Roslyn</Value>
+      <Value Profile="(Default)" />
     </Setting>
   </Settings>
 </SettingsFile>
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Packages.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Packages.cs
@@ -2,7 +2,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using NuGet.Versioning;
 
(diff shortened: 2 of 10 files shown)


jakubch1:
@RikkiGibson  @JoeRobich could you please take a look on this PR?

RikkiGibson:
Could you please link an example insertion that was created by passing an argument to the new command line option?

jakubch1:
@RikkiGibson  I've just tested it locally:
PS D:\roslyn-tools> D:\roslyn-tools\artifacts\bin\RIT\Debug\net46\RIT.exe --skipcorextpackages "package1,package2,package3"
Processing args:
--skipcorextpackages
package1,package2,package3

Is there a way to use it before merging to main?

RikkiGibson:
You should be able to make an azdo token and use the tool to open a PR against the VS repo under your name. If you run it without providing an auth argument it should spit out instructions on how to generate the token.

jakubch1:
This is link to PR: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/329196?_a=files&path=%2F.corext%2FConfigs%2Fdefault.config

Generated by:

D:\roslyn-tools\artifacts\bin\RIT\Debug\net46\RIT.exe /in="VS Code Coverage" /bn=main /bq=MSTest.CI /vsbn=main /ic=true /id=true /qv=true /ua=false /uc=true /username=*** /password="***" /ci=4b5bb1f2-c557-40c4-b166-ad368c5378a8 /tp="[Auto Insertion]" /ac=true /dpr=false --skipcorextpackages "Microsoft.VisualStudio.Coverage.Interprocess,Microsoft.VisualStudio.Coverage.Instrumentation,Microsoft.VisualStudio.Coverage.Core,Microsoft.VisualStudio.TraceDataCollector"

Updates of those 4 packages was correctly excluded

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Packages.cs
```diff
@@ -48,9 +51,8 @@ public OutdatedPackageException(string message, PackageInfo package, NuGetVersio
 
                 var package = PackageInfo.ParsePackageFileName(fileName);
 
-                if (package.IsRoslynToolsetCompiler)
+                if (packagesToBeIgnored.Any(p => p == package.PackageName))
```
JoeRobich, line 54: I think we would want to keep the separate check for RoslynToolsetCompiler. I know the package is the default value. But, if Roslyn wanted to adopt this setting in the future, we would have to remember to include the VS.Tools.Roslyn package name.

src/RoslynInsertionTool/RoslynInsertionTool.Commandline/Settings.settings
```diff
@@ -71,5 +71,8 @@
     <Setting Name="VisualStudioRepoAzdoUri" Type="System.String" Scope="Application">
       <Value Profile="(Default)">https://devdiv.visualstudio.com/DefaultCollection</Value>
     </Setting>
+    <Setting Name="SkipCoreXTPackages" Type="System.String" Scope="Application">
+      <Value Profile="(Default)">VS.Tools.Roslyn</Value>
```
JoeRobich, line 75: I don't think we should have a default value for this option. Here or in the app.config.

JoeRobich (commented):
Thanks this looks like a good option to provide. My only requested change is that we preserve the separate check for the Roslyn compiler package.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.Packages.cs
jakubch1: changed

src/RoslynInsertionTool/RoslynInsertionTool.Commandline/Settings.settings
jakubch1: changed

jakubch1:
@JoeRobich could you please merge it if there is no more comments?

JoeRobich (approved):

PR #1008, opened 2021-06-21
Add option of reviewerGUID instead of defaulting to MLInfraSwat

### LINKED ISSUE #996: PRs shouldn't get auto assigned to MLInfraSwat - instead take a cmd line parameter
These shouldn't be hardcoded and instead should be part of the release pipeline of each team - this way each team can set it to the appropriate GUID and searching through their PRs would be easier as well.

https://github.com/dotnet/roslyn-tools/blob/24bb13d9ebe6e50cd976529c8b7d625954d7bfe2/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs#L27

dmonroym:
Fixes #996

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
@@ -24,7 +24,7 @@ namespace Roslyn.Insertion
         public static readonly Guid VSRepoId = new Guid("a290117c-5a8a-40f7-bc2c-f14dbe3acf6d");
         //Easiest way to get these GUIDs is to create a PR search in AzDo
         //You'll get something like https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequests?_a=active&createdBy=GUID-here
-        public static readonly Guid MLInfraSwatUserId = new Guid("6c25b447-1d90-4840-8fde-d8b22cb8733e");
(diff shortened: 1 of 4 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.cs
```diff
@@ -24,7 +24,7 @@ public static partial class RoslynInsertionTool
         public static readonly Guid VSRepoId = new Guid("a290117c-5a8a-40f7-bc2c-f14dbe3acf6d");
         //Easiest way to get these GUIDs is to create a PR search in AzDo
         //You'll get something like https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequests?_a=active&createdBy=GUID-here
-        public static readonly Guid MLInfraSwatUserId = new Guid("6c25b447-1d90-4840-8fde-d8b22cb8733e");
+        public static readonly Guid ReviewerUserId = new Guid($"{Options.ReviewerGUID}");
```
RikkiGibson, line 27: If it is practical, I think it would be best to delete this field. Also, it seems like this may have wrong semantics, since this initializer will run before we set the `Options` in `PerformInsertionAsync` below.

dmonroym: Yes, this was done lazily and I sent the PR before testing just in case I needed to run checks but will be cleaning that up and since I can test locally I will abandon this now until it's ready.

PR #1018, opened 2021-06-25
Generate Roslyn tags for dev17 releases

JoeRobich:
Also, supports locating Roslyn builds in dnceng

### DIFF
--- a/src/CreateTagsForVSRelease/AzDOConnection.cs
+++ b/src/CreateTagsForVSRelease/AzDOConnection.cs
@@ -9,7 +9,7 @@ using Azure.Security.KeyVault.Secrets;
 
 namespace CreateTagsForVSRelease
 {
-    internal sealed class AzDOConnection: IDisposable
+    internal sealed class AzDOConnection : IDisposable
     {
         private bool _disposed = false;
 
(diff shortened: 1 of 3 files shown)


src/CreateTagsForVSRelease/AzDOConnection.cs
```diff
@@ -0,0 +1,51 @@
...
+using Microsoft.VisualStudio.Services.Common;
+using Microsoft.VisualStudio.Services.WebApi;
+using System;
+using System.Net.Http;
+using System.Net;
+using Azure.Security.KeyVault.Secrets;
+
+namespace CreateTagsForVSRelease
+{
+    internal sealed class AzDOConnection: IDisposable
```
RikkiGibson, line 12: nit

```suggestion
    internal sealed class AzDOConnection : IDisposable
```

RikkiGibson:
Did you push any tags with this version of the tool? Or does it need to be merged before we can do that?

JoeRobich:
> Did you push any tags with this version of the tool? Or does it need to be merged before we can do that?

I pushed the dev17 and dev16.11p2 tags with it yesterday

RikkiGibson (approved):

PR #1025, opened 2021-06-30
Automatically update LegacyProjects.props

### LINKED ISSUE #1024: Add support for LegacyProjects.props
Currently insertions to VS main are broken (require manual editing) because the insertion tool hasn't accounted for the new LegacyProjects.props. We need to fix this as soon as reasonably possible in order to get our automation healthy again.

RikkiGibson:
Resolves #1024 

PTAL @JoeRobich @dotnet/roslyn-infrastructure

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
@@ -20,15 +20,22 @@ namespace Roslyn.Insertion
         private static HashSet<string> dirtyFiles;
 
         private const string DefaultConfigPath = ".corext/Configs/default.config";
+        private const string LegacyProjectPropsPath = "src/ConfigData/Packages/LegacyProjects.props";
         private const string ComponentsJsonPath = ".corext/Configs/components.json";
+
         private readonly string _defaultConfigOriginal;
+        private readonly string _legacyPropsOriginal;
 
         public XDocument ConfigDocument { get; }
+        public XDocument LegacyPropsDocument { get; }
 
-        public CoreXT(string configOriginalText)
+        private CoreXT(string configOriginalText, string legacyOriginalText)
         {
             _defaultConfigOriginal = configOriginalText;
             ConfigDocument = XDocument.Parse(configOriginalText, LoadOptions.None);
+
+            _legacyPropsOriginal = legacyOriginalText;
+            LegacyPropsDocument = XDocument.Parse(legacyOriginalText, LoadOptions.None);
         }
 
         public static async Task<CoreXT> Load(GitHttpClient gitClient, string commitId)
@@ -41,8 +48,14 @@ namespace Roslyn.Insertion
                 DefaultConfigPath,
                 download: true,
                 versionDescriptor: vsBranch);
+            using var legacyPropsStream = await gitClient.GetItemContentAsync(
+                vsRepoId,
+                LegacyProjectPropsPath,
+                download: true,
+                versionDescriptor: vsBranch);
 
             var defaultConfigOriginal = await new StreamReader(defaultConfigStream).ReadToEndAsync();
+            var legacyPropsOriginal = await new StreamReader(legacyPropsStream).ReadToEndAsync();
 
             ComponentToFileMap = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             ComponentFileToDocumentMap = new Dictionary<string, (string, JObject)>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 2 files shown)


RikkiGibson:
Example insertion: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/334999

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -41,21 +48,43 @@ public static async Task<CoreXT> Load(GitHttpClient gitClient, string commitId)
                 DefaultConfigPath,
                 download: true,
                 versionDescriptor: vsBranch);
+            using var legacyPropsStream = await gitClient.GetItemContentAsync(
```
dibarbet, line 51: do we need to handle cases where the legacy props file doesn't exist (e.g. older branches of VS?

dibarbet (commented):
from reading the response to your original email, the micro build task was updated appropriately.  Do you happen to know what is blocking us from using the micro build VS Insertion task?

RikkiGibson:
IIRC, microbuild insertion tool isn't capable of updating the toolset. Also, RIT has a number of quality of life features that I think would be too bad to lose out on. adding @jmarolf who has experience using the microbuild tool IIRC.

jmarolf:
> IIRC, microbuild insertion tool isn't capable of updating the toolset

Haven't re-evaluated it in over a year. It may meet our needs now but at the time it only supported a small set of the files we _technically_ need to update. That set of files may have changed and/or been removed in the intervening years.

RikkiGibson:
Insertion into an d16.8 (using a main-vs-deps build to ensure we observe the difference): https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/335032?_a=files

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
RikkiGibson: Fixed. Thanks for catching this.

JoeRobich (approved):

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -41,21 +51,57 @@ public static async Task<CoreXT> Load(GitHttpClient gitClient, string commitId)
...
+            {
+                using var legacyPropsStream = await gitClient.GetItemContentAsync(
+                    vsRepoId,
+                    LegacyProjectPropsPath,
+                    download: true,
+                    versionDescriptor: vsBranch);
+
+                legacyPropsOriginal = await new StreamReader(legacyPropsStream).ReadToEndAsync();
+            }
+            catch (VssServiceException)
```
dibarbet, line 67: is there a particular property on the exception / exception subtype we could look for that means the file did not exist (versus some other random error

dibarbet (approved):

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
dibarbet: or at least logging the exception I think would be useful, in case we see missing changes to that legacy props file that shouldn't be missing

RikkiGibson: Unfortunately, I couldn't find a way to more specifically identify a "not found", versus a network failure, for example.

RikkiGibson: I added some log statements in d518717. The output looks like this

```
New Insertion Into rel/d16.8 Started

Verifying given authentication for https://devdiv.visualstudio.com/DefaultCollection
Creating VisualStudioRepoConnection object from https://devdiv.visualstudio.com/DefaultCollection
Verification succeeded for https://devdiv.visualstudio.com/DefaultCollection
Creating ComponentBuildConnection object from https://dev.azure.com/dnceng/
Verifying given authentication for https://dev.azure.com/dnceng/
Verification succeeded for https://dev.azure.com/dnceng/
Get Latest Passed Component Build
Getting latest passing build for project internal, queue dotnet-roslyn CI, and branch main-vs-deps
Found build number 20210630.1
Downloading artifacts to C:\Users\rikki\AppData\Local\Temp\Roslynmain-vs-deps\VSSetup
Artifact download took 20 seconds
Unable to load LegacyProjects.props. It will not be updated in this insertion.
TF401174: The item 'src/ConfigData/Packages/LegacyProjects.props' could not be found in the repository 'VS' at the version specified by '<Commit: 3a4f23a782c88da2528f578230ad991f365a2d56 >' (resolved to commit '3a4f23a782c88da2528f578230ad991f365a2d56')
...
```

jmarolf:
@dibarbet 
Took a look at the tasks source (internal link [here](https://devdiv.visualstudio.com/Engineering/_git/MicroBuild?path=%2Fsrc%2FTasks%2FInsertVsPayload%2Fplugin.ps1&_a=contents&version=GBmain)). They do not support updating opt-prof data. We could try to somehow do this in a powershell script that the microbuild task calls, but I don't see how that is much different from having a custom insertion tool if we need a script to do critical parts of the logic.

dibarbet (approved):

PR #1032, opened 2021-07-06
Pass reviewer to RIT in CreateDummyPR script

JoeRobich:


RikkiGibson (approved):
This may fix the problem, so would be happy to take it as is. We might also want to check on why the tool is failing now and not before--is the application passing an empty string for the reviewerGUID or something? If so then perhaps empty string should be replaced with null before we call through to the client, for example.

JoeRobich:
I'll come back to this when I have time

RikkiGibson:
In case it helps, I'm fairly confident that the break was introduced in this range: https://github.com/dotnet/roslyn-tools/compare/7f8a530393ba2b885018e2a5449285eefe4aebd0...326524d65f9835677e4a701a1c4125e18ef1c0c8

JoeRobich:
Likely issue is https://github.com/dotnet/roslyn-tools/compare/7f8a530393ba2b885018e2a5449285eefe4aebd0...326524d65f9835677e4a701a1c4125e18ef1c0c8#diff-75bc1974ca9b319674ec6874dfd86e930ae879ddc57f842b4a6464842a93d2ebR62. In other locations we use the build queuer id when reviewer guid is null.

PR #1073, opened 2021-08-18
Update Microsoft.VisualStudio.Threading resolution path

JoeRobich:
In VS 17.0 Preview 3 the path to the Microsoft.VisualStudio.Threading library has changed. Without the updated path the 15.8 version of this library is discovered.

### DIFF
--- a/src/VSIXExpInstaller/Program.cs
+++ b/src/VSIXExpInstaller/Program.cs
@@ -157,7 +157,7 @@ namespace VsixExpInstaller
 
                 var assemblyResolutionPaths = new string[] {
                     // Microsoft.VisualStudio.Threading, Version=17.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
-                    Path.Combine(vsInstallDir, @"Common7\IDE\PublicAssemblies\Microsoft.VisualStudio.Threading.17.0"),
+                    Path.Combine(vsInstallDir, @"Common7\IDE\PublicAssemblies\Microsoft.VisualStudio.Threading.17.x"),
                     // Newtonsoft.Json, Version=12.0.0.2, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed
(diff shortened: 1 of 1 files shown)


dibarbet (approved):

sharwell:
@AArnott I wouldn't complain if the assembly in Common7\\IDE was a usable 17.0 version...

AArnott:
> expecting tools to dig around for a better one.

I *don't* expect tools to dig around for this. I don't want tools to reuse this. They should ship their own dependencies. It is precisely tools "finding" the one in PublicAssemblies that prevents us from upgrading it, since changing the assembly version would break those tools that haven't yet updated their binding redirects.

As you've apparently gone to trouble to "find" the new 17.0 version, I wonder are you either a .NET Core tool that doesn't care about assembly versions, or do you load the DLL in an appdomain that also reuses devenv.exe.config in order to also stay current on the binding redirects necessary to consume the dll?

sharwell:
> As you've apparently gone to trouble to "find" the new 17.0 version, I wonder are you either a .NET Core tool that doesn't care about assembly versions, or do you load the DLL in an appdomain that also reuses devenv.exe.config in order to also stay current on the binding redirects necessary to consume the dll?

It's a command line tool to automate VSIX installation for integration tests. Most of the runtime dependencies don't have redist licenses so we have to dynamically bind to them in the installation directory.

AArnott:
Are you then using reflection such that binding redirects aren't required?
FWIW, vs-threading _is_ redistributable.

sharwell:
> Are you then using reflection such that binding redirects aren't required?
> FWIW, vs-threading is redistributable.

We compile against reference assemblies, and generate the necessary binding redirects. The problem occurred when we had an assembly binding redirect for MS.VS.Threading that expected to load 17.0, but at runtime the resolve handler returned 15.8.

AArnott:
I wonder if you should do all your work in a custom appdomain that is configured to use devenv.exe.config. Then you'll get the right binding redirects and the right DLLs to match. And it could run on multiple versions of VS.

PR #1097, opened 2021-09-17
Update the version to avoid restoring System.Net.Http

Cosifne:
Issue: https://devdiv.visualstudio.com/DevDiv/_componentGovernance/112189/alert/2974928?typeId=6256257

The root reason is many of the packages are referencing NetStandard.Library 1.6.0, which is referencing System.Net.Http 4.1.0
For example:
image

Updating the version of this package would resolve the problem.

For other packages, which the updated version still referencing NetStandard.Library 1.6.0, explicitly referencing System.Net.Http 4.3.4

### DIFF
--- a/src/CreateTagsForVSRelease/CreateTagsForVSRelease.csproj
+++ b/src/CreateTagsForVSRelease/CreateTagsForVSRelease.csproj
@@ -14,6 +14,7 @@
     <PackageReference Include="LibGit2Sharp" Version="0.26.2" />
     <PackageReference Include="Microsoft.TeamFoundationServer.Client" Version="16.153.0" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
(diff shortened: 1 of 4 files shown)


src/CreateTagsForVSRelease/CreateTagsForVSRelease.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageReference Include="LibGit2Sharp" Version="0.26.2" />
     <PackageReference Include="Microsoft.TeamFoundationServer.Client" Version="16.153.0" />
```
Cosifne, line 15: I really like to move all these versions to Versions.props.
However, because all of these projects are using different TFM because of the long history, change starts to blow up.
In future, if we have a chance to sync all projects to Net6.0 then this would be feasible

PR #1119, opened 2021-11-01
Fix create GitHub merge pipeline

JoeRobich:


RikkiGibson (approved):
We don't have some other task doing the same thing here right?

JoeRobich:
@RikkiGibson Not sure what you mean. We do a GH PR dry-run as part of PR CI but it fails so consistently due to rate limiting that I have gotten out of the habit of checking it.

JoeRobich:
Not sure what about moving from netcoreapp2.0 to netcoreapp3.1 broke this other than maybe the NET 2.1 runtime is installed on `ubuntu-latest` but a NET 3.1 runtime isn't.

PR #1139, opened 2021-11-30
Recurse through Packages subdirectories

tmeschter:
The Roslyn Insertion Tool already tries to update `PackageReference` items in files under src/ConfigData/Packages in the VS repo. However, it only looks at files within that directory, and does not examine files in subdirectories. This means it fails to find and update Packages/Wizard/Packages.props and Packages/ProjectSystem/Managed.props when inserting builds of dotnet/project-system.

The fix is to retrieve **all** items under src/ConfigData/Packages when building a mapping from the package name to the .props file with the corresponding `PackageReference`, not just the top-level items.


RikkiGibson (approved):
Change LGTM, would be nice to get a test insertion created with this version of the tool before merging or at least give a heads up to infraswat and have a plan to revert if things go sideways.

JoeRobich (approved):

tmeschter:
@RikkiGibson It's been years since I touched this. Any guidance on how to create a test insertion with this version?

MiYanni (approved):

JoeRobich:
Test Insertion: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/367305 (microsoft)

PR #1158, opened 2022-01-10
Update config.xml

tmeschter:
Update config.xml to merge from dotnet/project-system dev17.1.x to main.


MiYanni (approved):
This is fine and can be done independent of my comments on https://github.com/dotnet/project-system/pull/7843. This config change can happen at any time since once it is merged, it'll always compare the history of the branches when it runs.

PR #1160, opened 2022-01-20
Remove 6.0.2xx from linker merges

agocke:
We were planning to put regular changes into that branch, but now it's just a backporting release branch, so it doesn't need merge configuration.


sbomer (approved):
Thank you!

agocke:
@RikkiGibson Looks like I need a review from someone else with merge permission, mind signing off? :)

dibarbet (approved):

RikkiGibson:
Thanks for taking care of it @dibarbet :)

Should we adjust some settings so that collaborators can merge without review?

PR #1181, opened 2022-03-11
Delete release/dev17.1-vs-deps from merge PR flow

allisonchou:
We've deleted release/dev17.1-vs-deps in Roslyn, so any merge PRs to/from the branch are unnecessary.

### DIFF
--- a/src/GitHubCreateMergePRs/config.xml
+++ b/src/GitHubCreateMergePRs/config.xml
@@ -12,11 +12,10 @@
     <merge from="release/dev17.0-vs-deps" to="release/dev17.1" />
 
     <!-- Roslyn branches (from non-vs-deps to vs-deps) -->
-    <merge from="release/dev17.1" to="release/dev17.1-vs-deps" />
     <merge from="main" to="main-vs-deps" />
 
     <!-- Roslyn release branches (flowing between releases) -->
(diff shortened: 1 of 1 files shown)


RikkiGibson (approved):

src/GitHubCreateMergePRs/config.xml
```diff
@@ -12,11 +12,10 @@
...
 
     <!-- Roslyn branches (from non-vs-deps to vs-deps) -->
-    <merge from="release/dev17.1" to="release/dev17.1-vs-deps" />
     <merge from="main" to="main-vs-deps" />
 
     <!-- Roslyn release branches (flowing between releases) -->
-    <merge from="release/dev17.1-vs-deps" to="release/dev17.2" />
+    <merge from="release/dev17.1" to="release/dev17.2" />
     <merge from="release/dev17.2" to="main" />
     <merge from="main" to="release/dev17.3" />
```
RikkiGibson, line 20: incidentally, should 17.3 branch contain any vs-deps bits here? Are we hoping to remove -vs-deps again before it's time to start inserting 17.3 anyway?

allisonchou: Oh, great point - I think it might be good to set up that flow just in case. Will open a follow-up PR since this one already auto-merged

PR #1182, opened 2022-03-11
Flow main-vs-deps to release/dev17.3

allisonchou:
@RikkiGibson brought up a good point that main-vs-deps should probably flow into release/dev17.3, at least for now, since we don't know whether we'll still need the main-vs-deps branch once 17.3 starts shipping.


RikkiGibson (approved):
lgtm assuming we are fine with the need for internal bits to work in the 17.3 branch.

PR #1185, opened 2022-03-23
Move VSBranchInfo to the dotnet tool, and add Razor support

davidwengier:
Thank you for setting this up Joey


JoeRobich (approved):

src/dotnet-roslyn/VS/VSBranchInfo.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            throw new Exception($"Resource data value '{resourceData}' was not in expected format.");
+        }
+    }
+
+    private static void WriteHeader(string branch)
+    {
+        Console.ForegroundColor = ConsoleColor.Cyan;
+        Console.WriteLine($"{branch}:");
+        Console.ResetColor();
```
JoeRobich, line 135: Could use the `logger` for writing your output, although you would lose direct control over the color without additional work.

davidwengier: I thought about that, but then when writing out the info, I would never know if the user can actually see it, or if they passed `-v minimal` and are wondering why the app doesn't do anything :)

JoeRobich: you never can tell with users =) anyhoo, looks good to me

PR #1189, opened 2022-03-25
Reduce, reuse, recycle

davidwengier:
It's a pity there aren't any tests, but I'm at least 10% sure this will all still work 😁


davidwengier:
Updated after your rename @JoeRobich, if you want to take a look?

JoeRobich (approved):

PR #1210, opened 2022-04-25
update F# merges

brettfo:


vzarytovskii (approved):

src/GitHubCreateMergePRs/config.xml
```diff
@@ -55,9 +55,10 @@
     <merge from="release/dev16.11" to="release/dev17.0" />
     <merge from="release/dev17.0" to="release/dev17.1" />
     <merge from="release/dev17.1" to="release/dev17.2" />
+    <merge from="release/dev17.2" to="release/dev17.3" />
 
     <!-- regular branch flow -->
-    <!--<merge from="main" to="release/dev17.2" />-->
```
JoeRobich, line 60: We should consider excluding this file from the various CI checks. as in https://github.com/dotnet/roslyn/blob/main/azure-pipelines.yml#L18-L21=

JoeRobich (approved):

src/GitHubCreateMergePRs/config.xml
JoeRobich: Opened https://github.com/dotnet/roslyn-tools/pull/1211 to exclude this file

PR #1251, opened 2022-07-26
Add VS insertion PR tagger

allisonchou:
- Implements a PR tagger that creates an issue with the PRs that were inserted in a given VS build.
- Currently works with Roslyn and Razor but can likely be extended to other repos as well.
- After this is merged, we should be able to trigger runs automatically via AzDo pipelines.

### DIFF
--- a/src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
+++ b/src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
@@ -1,0 +1,48 @@
+// Licensed to the.NET Foundation under one or more agreements.
+// The.NET Foundation licenses this file to you under the MIT license.
+// See the License.txt file in the project root for more information.
+
+namespace Microsoft.RoslynTools.Commands;
+
+using System.CommandLine.Invocation;
+using System.CommandLine;
+
+internal static class PRTaggerCommand
+{
+    private static readonly PRTaggerCommandDefaultHandler s_prTaggerCommandHandler = new();
+
+    internal static readonly Option<string> VSBuild = new(new[] { "--build", "-b" }, "VS build number") { IsRequired = true };
+    internal static readonly Option<string> CommitId = new(new[] { "--commit", "-c" }, "VS build commit") { IsRequired = true };
+    internal static readonly Option<string> GitHubUsername = new(new[] { "--username", "-u" }, "GitHub username") { IsRequired = true };
+    internal static readonly Option<string> GitHubPassword = new(new[] { "--password", "-p" }, "GitHub password") { IsRequired = true };
+
+    public static Symbol GetCommand()
+    {
+        var command = new Command("pr-tagger", "Tags PRs inserted in a given VS build.")
+        {
+            CommitId,
+            GitHubUsername,
+            GitHubPassword,
+            CommonOptions.DevDivAzDOTokenOption,
+        };
+
+        command.Handler = s_prTaggerCommandHandler;
+        return command;
+    }
+
+    private class PRTaggerCommandDefaultHandler : ICommandHandler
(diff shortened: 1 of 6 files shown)


davidwengier (approved):
Nice work! Love that we can start re-using some of the infrastructure built up in this tool.

src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
```diff
@@ -0,0 +1,57 @@
...
+internal static class PRTaggerCommand
+{
+    private static readonly PRTaggerCommandDefaultHandler s_prTaggerCommandHandler = new();
+
+    private static readonly Option<string> ProductName = new(new[] { "--product-name", "-p" }, "Name of product (e.g. 'roslyn' or 'razor')") { IsRequired = true };
+    private static readonly Option<string> VSBuild = new(new[] { "--build", "-b" }, "VS build number") { IsRequired = true };
+    private static readonly Option<string> CommitId = new(new[] { "--commit", "-c" }, "VS build commit") { IsRequired = true };
+    private static readonly Option<string> GitHubToken = new(new[] { "--github-token" }, "Token used to authenticate GitHub.") { IsRequired = true };
+    private static readonly Option<string> DevDivAzDOToken = new(new[] { "--devdiv-azdo-token" }, "Token used to authenticate to DevDiv Azure DevOps.") { IsRequired = true };
+    private static readonly Option<string> DncEngAzDOToken = new(new[] { "--dnceng-azdo-token" }, "Token used to authenticate to DncEng Azure DevOps.") { IsRequired = true };
```
JoeRobich, line 20: I think we should handle these tokens the same way `VSBranchInfo` does - https://github.com/dotnet/roslyn-tools/blob/main/src/dotnet-roslyn-tools/Commands/VSBranchInfoCommand.cs#L43. Add the token options from CommonOptions and then parse the setting from the `ParseResult`. Invalid/missing tokens could be checked up front or by simply allow the tagging to throw an exception with authorization issues occur.

src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using System.CommandLine.Invocation;
+using System.CommandLine;
+using Microsoft.RoslynTools.Authentication;
+
+internal static class PRTaggerCommand
+{
+    private static readonly PRTaggerCommandDefaultHandler s_prTaggerCommandHandler = new();
+
+    private static readonly Option<string> ProductName = new(new[] { "--product-name", "-p" }, "Name of product (e.g. 'roslyn' or 'razor')") { IsRequired = true };
```
JoeRobich, line 15: We can use `.FromAmong(string[])` to ensure a valid product name. See https://docs.microsoft.com/en-us/dotnet/standard/commandline/define-commands#list-valid-argument-values

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -0,0 +1,205 @@
...
+            new
+            {
+                title = $"[Automated] PRs inserted in VS build {vsBuildNumber}",
+                body = issueBody,
+                labels = new string[] { "vs-insertion" }
+            }));
+
+        if (!response.IsSuccessStatusCode)
+        {
+            logger.LogInformation($"Issue creation failed with status code: {response.StatusCode}");
```
JoeRobich, line 198: Maybe LogWarning or LogError.
```suggestion
            logger.LogError($"Issue creation failed with status code: {response.StatusCode}");
```

JoeRobich (approved):
Nice!

PR #1252, opened 2022-07-26
Update merge PR flow for 17.4 P1 snap

allisonchou:


dibarbet (approved):
Just looking at versions.props I think we should try merging main-vs-deps back to main.  But we can do that at a later point.

src/GitHubCreateMergePRs/config.xml
```diff
@@ -15,10 +15,13 @@
     <merge from="release/dev17.2" to="release/dev17.3" />
 
     <!-- Roslyn release branches (flowing between releases) -->
-    <merge from="release/dev17.3" to="main" />
+    <merge from="release/dev17.3" to="release/dev17.4" />
+    <merge from="release/dev17.3-vs-deps" to="release/dev17.4-vs-deps" />
+    <merge from="release/dev17.4" to="main" />
```
JoeRobich, line 21: Looks like we need a dev17.4-vs-deps to main-vs-deps merge until we are able to remove -vs-deps branches.

PR #1262, opened 2022-08-03
[PR tagger] Add F# support + clone repos directly in code

allisonchou:
- Adds PR tagger support for F#
- Instead of having the release pipeline download the product repo, we now clone the repo directly in the code. This enables us to have a singular pipeline do the work.


src/dotnet-roslyn-tools/Products/FSharp.cs
```diff
@@ -0,0 +1,24 @@
+// Licensed to the.NET Foundation under one or more agreements.
+// The.NET Foundation licenses this file to you under the MIT license.
+// See the License.txt file in the project root for more information.
+
+namespace Microsoft.RoslynTools.Products;
+
+internal class FSharp : IProduct
```
dibarbet, line 7: would we prefer these to be configurable via input parameters to the tool?  That way if anyone wants a new repo we can just have a new pipeline without needing to update the tool?

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -92,6 +93,17 @@ internal class PRTagger
 
         logger.LogInformation($"Finding PRs between {productInfo.Name} commit SHAs {previousProductCommitSha} and {currentProductCommitSha}.");
 
+        string? productRepoPath = null;
+        try
+        {
+            productRepoPath = Repository.Clone(sourceUrl: productInfo.RepoBaseUrl, workdirPath: Environment.CurrentDirectory + productInfo.Name);
```
dibarbet, line 99: should this be part of the pipeline definition?

src/dotnet-roslyn-tools/Products/FSharp.cs
dibarbet: these might have been already answered in previous reviews, let me know if so

src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
```diff
@@ -13,8 +13,7 @@ internal static class PRTaggerCommand
     private static readonly PRTaggerCommandDefaultHandler s_prTaggerCommandHandler = new();
 
     private static readonly Option<string> ProductName = new Option<string>(new[] { "--product-name", "-n" }, "Name of product (e.g. 'roslyn' or 'razor')") { IsRequired = true }
-        .FromAmong("roslyn", "razor");
-    private static readonly Option<string> ProductRepoPath = new(new[] { "--repo-path", "-p" }, "Path to product repo") { IsRequired = true };
+        .FromAmong("roslyn", "razor", "fsharp");
```
davidwengier, line 16: I think I have some code for `AllProducts` or something in the vsbranchinfo command.. might be something to generalize? eg, pretty sure this can work for typescript too, with no changes

src/dotnet-roslyn-tools/Products/FSharp.cs
davidwengier: I think it makes sense that each command could have extra optional parameters to let people pass in the bits that command needs, rather than having to have a pre-defined product. Makes a lot of sense for calling these tools from a CI pipeline.

For using them on the command line directly, which is what I do for VS Mac insertions, having pre-defined products is really nice too :)

davidwengier (approved):

src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
JoeRobich: I think I would like to see this become non-required. Allow it to be passed to limit for testing or other purposes, but the default behavior would be to spin through the products and look for changes.

src/dotnet-roslyn-tools/Products/FSharp.cs
JoeRobich: I'm not against have other ways to specify a product. If you have a proposal, please open an issue and we can consider it. Personally, I hope more of the tooling becomes multi-product aware.

allisonchou: So I just updated the code to do both - we have a set of default products (for more ease of use on command line), but instead users can also pass in their own arguments. Hopefully this approach makes everyone happy 😄

src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
allisonchou: Yep, just changed it to be non-required 😄

src/dotnet-roslyn-tools/Products/FSharp.cs
davidwengier: Never! I refuse to be happy!

allisonchou:
Changed the approach of the PR tagger again based on @JoeRobich's suggestion, which I think is a sound approach:

> ... a single pipeline will work for every team who has a Product registered in roslyn-tools. Each time it is triggered it will loop through all the products that it knows about, checking for whether the component has been updated. If it has, then it will clone and tag the PRs appropriately. Then move on to check the remaining products.

The updated approach just loops for all the registered products, which enables us to (1) only use one pipeline and (2) integrate better with the rest of roslyn-tools.

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -50,92 +46,97 @@ internal class PRTagger
...
-            return -1;
-        }
+            // Get associated product build for current and previous VS commit SHAs
+            var currentBuild = await TryGetBuildNumberForReleaseAsync(product.ComponentJsonFileName, product.ComponentName, vsCommitSha, devdivConnection).ConfigureAwait(false);
+            var previousBuild = await TryGetBuildNumberForReleaseAsync(product.ComponentJsonFileName, product.ComponentName, previousVsCommitSha, devdivConnection).ConfigureAwait(false);
 
-        // Get associated product build for current and previous VS commit SHAs
-        var currentBuild = await TryGetBuildNumberForReleaseAsync(productInfo, vsCommitSha, devdivConnection).ConfigureAwait(false);
-        var previousBuild = await TryGetBuildNumberForReleaseAsync(productInfo, previousVsCommitSha, devdivConnection).ConfigureAwait(false);
+            var gitHubRepoName = product.RepoBaseUrl.Split('/').Last();
```
davidwengier, line 55: Should check if this actually starts with "github.com", or have a boolean flag on Product that controls whether PRs will automatically be tagged. 

eg TypeScript has a `Product` but its not a github repo

allisonchou: Good call, thanks!

JoeRobich (commented):
👍

PR #1281, opened 2022-09-30
Support updating a common property that specifies package versions

jasonmalinowski:


src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -135,14 +142,32 @@ public List<GitChange> SaveComponents()
...
                 .Where(p => p.Attribute("Include")?.Value == packageInfo.PackageName || p.Attribute("Update")?.Value == packageInfo.PackageName)
                 .Select(x => x.Attribute("Version")).SingleOrDefault();
+
+            if (attribute is null)
+                return null;
+
+            // If the attribute is a property substitution, then we'll see if we actually can find the property itself
+            if (attribute.Value.StartsWith("$(") && attribute.Value.EndsWith(")"))
+            {
+                var propertyName = attribute.Value.Substring(2, attribute.Value.Length - 3);
```
dibarbet, line 163: are we able to use ranges in this project?

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -183,8 +208,11 @@ public void UpdatePackageVersion(PackageInfo packageInfo)
                     (string? original, XDocument? document) pair = (null, null);
                     if (PackagePropFileToDocumentMap.TryGetValue(file, out pair!))
                     {
-                        var propsVersionAttribute = GetVersionAttributeInPropsFileOpt(pair.document!, packageInfo);
-                        propsVersionAttribute?.SetValue(packageInfo.Version.ToString());
+                        var versionObject = GetVersionAttributeOrElementInPropsFile(pair.document!, packageInfo);
+                        if (versionObject is XAttribute attribute)
+                            attribute.SetValue(packageInfo.Version.ToString());
+                        else if (versionObject is XElement element)
+                            element.SetValue(packageInfo.Version.ToString());
```
dibarbet, line 215: should we do something (throw / log) if neither is true?

dibarbet (approved):

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
jasonmalinowski: We're targeting net46 so no idea. I'd also have to learn how to use ranges for this. :smile:

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
jasonmalinowski: Ah, probably. I don't understand why the null ?. was there before, but you're right this makes it worse if the function were to return non-null but some other type.

arkalyanms (approved):

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
jasonmalinowski: Added a throw.

PR #1297, opened 2022-11-21
Allow create-release-tags to work for Razor

davidwengier:
Hopefully this won't break anything 🤷‍♂️

### DIFF
--- a/src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
+++ b/src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
@@ -21,7 +21,7 @@ internal class CreateReleaseTagsCommand
 
     public static Symbol GetCommand()
     {
-        var command = new Command("create-release-tags", "Generates git tags for VS releases in the Roslyn repo.")
+        var command = new Command("create-release-tags", "Generates git tags for VS releases in the repo.")
         {
             ProductOption,
             VerbosityOption,
--- a/src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
+++ b/src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
@@ -14,9 +14,10 @@ internal class CreateReleaseTagsCommand
 {
     private static readonly CreateReleaseTagsCommandDefaultHandler s_defaultHandler = new();
 
-    private static readonly string[] s_allProductNames = VSBranchInfo.AllProducts.Select(p => p.Name.ToLower()).ToArray();
+    // Filter the product to only those with git credentials, as we need to be able to commit to the repo to add tags
+    private static readonly string[] s_allProductNames = VSBranchInfo.AllProducts.Where(p => p.GitUserName.Length > 0).Select(p => p.Name.ToLower()).ToArray();
 
-    internal static readonly Option<string> ProductOption = new Option<string>(new[] { "--product", "-p" }, () => "roslyn", "Which product to get info for (roslyn or razor)").FromAmong(s_allProductNames);
+    internal static readonly Option<string> ProductOption = new Option<string>(new[] { "--product", "-p" }, () => "roslyn", "Which product to get info for").FromAmong(s_allProductNames);
 
     public static Symbol GetCommand()
     {
--- a/src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
+++ b/src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
@@ -4,6 +4,7 @@
 
 using System.CommandLine.Invocation;
 using System.CommandLine;
+using Microsoft.RoslynTools.VS;
 
 namespace Microsoft.RoslynTools.Commands;
 
@@ -13,10 +14,15 @@ internal class CreateReleaseTagsCommand
 {
     private static readonly CreateReleaseTagsCommandDefaultHandler s_defaultHandler = new();
 
+    private static readonly string[] s_allProductNames = VSBranchInfo.AllProducts.Select(p => p.Name.ToLower()).ToArray();
+
+    internal static readonly Option<string> ProductOption = new Option<string>(new[] { "--product", "-p" }, () => "roslyn", "Which product to get info for (roslyn or razor)").FromAmong(s_allProductNames);
+
     public static Symbol GetCommand()
     {
         var command = new Command("create-release-tags", "Generates git tags for VS releases in the Roslyn repo.")
         {
+            ProductOption,
             VerbosityOption,
             DevDivAzDOTokenOption,
             DncEngAzDOTokenOption
@@ -32,7 +38,9 @@ internal class CreateReleaseTagsCommand
             var logger = context.SetupLogging();
             var settings = context.ParseResult.LoadSettings(logger);
 
-            return await CreateReleaseTags.CreateReleaseTags.CreateReleaseTagsAsync(settings, logger);
+            var product = context.ParseResult.GetValueForOption(ProductOption)!;
+
+            return await CreateReleaseTags.CreateReleaseTags.CreateReleaseTagsAsync(product, settings, logger);
         }
     }
 }
--- a/src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
+++ b/src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
@@ -48,7 +48,7 @@ internal static class CreateReleaseTags
 
             logger.LogInformation("Tagging releases...");
 
-            foreach (var visualStudioRelease in visualStudioReleases.Reverse())
+            foreach (var visualStudioRelease in visualStudioReleases)
             {
                 var tagName = TryGetTagName(visualStudioRelease);
 
@@ -58,7 +58,7 @@ internal static class CreateReleaseTags
                     {
                         logger.LogInformation($"Tag '{tagName}' is missing.");
 
-                        RoslynBuildInformation? build = null;
+                        BuildInformation? build = null;
                         foreach (var connection in connections)
                         {
                             build = await TryGetBuildInfoForReleaseAsync(product, visualStudioRelease, devdivConnection, connection);
@@ -75,7 +75,7 @@ internal static class CreateReleaseTags
 
                             string message = $"Build Branch: {build.SourceBranch}\r\nInternal ID: {build.BuildId}\r\nInternal VS ID: {visualStudioRelease.BuildId}";
 
-                            repository.ApplyTag(tagName, build.CommitSha, new Signature("dotnet bot", "dotnet-bot@microsoft.com", when: visualStudioRelease.CreationTime), message);
+                            repository.ApplyTag(tagName, build.CommitSha, new Signature(product.GitUserName, product.GitEmail, when: visualStudioRelease.CreationTime), message);
                         }
                         else
                         {
@@ -105,7 +105,7 @@ internal static class CreateReleaseTags
         return 1;
     }
 
-    private static async Task<RoslynBuildInformation?> TryGetBuildInfoForReleaseAsync(IProduct product, VisualStudioVersion release, AzDOConnection vsConnection, AzDOConnection connection)
+    private static async Task<BuildInformation?> TryGetBuildInfoForReleaseAsync(IProduct product, VisualStudioVersion release, AzDOConnection vsConnection, AzDOConnection connection)
     {
         try
         {
@@ -127,7 +127,7 @@ internal static class CreateReleaseTags
                     {
                         if (!string.IsNullOrWhiteSpace(build.SourceBranch))
                         {
-                            return new RoslynBuildInformation(build.SourceVersion, build.SourceBranch, build.BuildNumber);
+                            return new BuildInformation(build.SourceVersion, build.SourceBranch, build.BuildNumber);
                         }
                     }
                 }
@@ -140,7 +140,7 @@ internal static class CreateReleaseTags
                 return null;
             }
 
-            var commitSha = await TryGetRoslynCommitShaFromBuildAsync(product, connection, buildNumber)
+            var commitSha = await TryGetCommitShaFromBuildAsync(product, connection, buildNumber)
                 ?? await TryGetRoslynCommitShaFromNuspecAsync(vsConnection, release);
             if (string.IsNullOrEmpty(commitSha))
             {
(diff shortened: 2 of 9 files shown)


davidwengier:
Ran this on the Razor repo, seemed to work fine: https://github.com/dotnet/razor/tags

davidwengier:
@RikkiGibson @dibarbet @ryzngard @jasonmalinowski anyone feel confident enough to review this?

src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
```diff
@@ -13,10 +14,15 @@ internal class CreateReleaseTagsCommand
 {
     private static readonly CreateReleaseTagsCommandDefaultHandler s_defaultHandler = new();
 
+    private static readonly string[] s_allProductNames = VSBranchInfo.AllProducts.Select(p => p.Name.ToLower()).ToArray();
+
+    internal static readonly Option<string> ProductOption = new Option<string>(new[] { "--product", "-p" }, () => "roslyn", "Which product to get info for (roslyn or razor)").FromAmong(s_allProductNames);
```
jasonmalinowski, line 19: So it looks like s_allProductNames will also include TypeScript and F# -- would those be expected to work or not? The hint text is unclear.

src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
```diff
@@ -32,7 +38,9 @@ public async Task<int> InvokeAsync(InvocationContext context)
             var logger = context.SetupLogging();
             var settings = context.ParseResult.LoadSettings(logger);
 
-            return await CreateReleaseTags.CreateReleaseTags.CreateReleaseTagsAsync(settings, logger);
+            var product = context.ParseResult.GetValueForOption(ProductOption)!;
```
jasonmalinowski, line 41: Why the ! here?

src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
```diff
@@ -32,7 +38,9 @@ public async Task<int> InvokeAsync(InvocationContext context)
...
+
+            return await CreateReleaseTags.CreateReleaseTags.CreateReleaseTagsAsync(product, settings, logger);
```
jasonmalinowski, line 43: Needs more CreateReleaseTags in this line.

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
```diff
@@ -2,42 +2,43 @@
...
-    public static async Task<int> CreateReleaseTagsAsync(RoslynToolsSettings settings, ILogger logger)
+    public static async Task<int> CreateReleaseTagsAsync(string productName, RoslynToolsSettings settings, ILogger logger)
     {
         try
         {
             using var devdivConnection = new AzDOConnection(settings.DevDivAzureDevOpsBaseUri, "DevDiv", settings.DevDivAzureDevOpsToken);
             using var dncengConnection = new AzDOConnection(settings.DncEngAzureDevOpsBaseUri, "internal", settings.DncEngAzureDevOpsToken);
 
-            var connections = new[] { devdivConnection, dncengConnection };
+            var product = VSBranchInfo.AllProducts.Single(p => p.Name.Equals(productName, StringComparison.OrdinalIgnoreCase));
```
jasonmalinowski, line 30: Should this mapping be happening in the command handler so this way this method just directly is given the IProduct?

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
```diff
@@ -47,43 +48,43 @@ public static async Task<int> CreateReleaseTagsAsync(RoslynToolsSettings setting
 
             logger.LogInformation("Tagging releases...");
 
-            foreach (var visualStudioRelease in visualStudioReleases)
+            foreach (var visualStudioRelease in visualStudioReleases.Reverse())
```
jasonmalinowski, line 51: This might need a comment for why the .Reverse() is there.

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
```diff
@@ -47,43 +48,43 @@ public static async Task<int> CreateReleaseTagsAsync(RoslynToolsSettings setting
...
+                if (tagName is not null)
                 {
-                    if (!existingTags.Any(t => t.FriendlyName == roslynTagName))
+                    if (!existingTags.Any(t => t.FriendlyName == tagName))
                     {
-                        logger.LogInformation($"Tag '{roslynTagName}' is missing.");
+                        logger.LogInformation($"Tag '{tagName}' is missing.");
 
-                        RoslynBuildInformation? roslynBuild = null;
+                        RoslynBuildInformation? build = null;
```
jasonmalinowski, line 61: Do we need to rename this type since it's not just Roslyn anymore?

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
```diff
@@ -47,43 +48,43 @@ public static async Task<int> CreateReleaseTagsAsync(RoslynToolsSettings setting
...
+                        if (build is not null)
                         {
-                            logger.LogInformation($"Tagging {roslynBuild.CommitSha} as '{roslynTagName}'.");
+                            logger.LogInformation($"Tagging {build.CommitSha} as '{tagName}'.");
 
-                            string message = $"Build Branch: {roslynBuild.SourceBranch}\r\nInternal ID: {roslynBuild.BuildId}\r\nInternal VS ID: {visualStudioRelease.BuildId}";
+                            string message = $"Build Branch: {build.SourceBranch}\r\nInternal ID: {build.BuildId}\r\nInternal VS ID: {visualStudioRelease.BuildId}";
 
-                            roslynRepository.ApplyTag(roslynTagName, roslynBuild.CommitSha, new Signature("dotnet bot", "dotnet-bot@microsoft.com", when: visualStudioRelease.CreationTime), message);
+                            repository.ApplyTag(tagName, build.CommitSha, new Signature("dotnet bot", "dotnet-bot@microsoft.com", when: visualStudioRelease.CreationTime), message);
```
jasonmalinowski, line 78: Considering earlier question around whether TypeScript/F# could use this -- would they use something other than dotnet-bot?

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
```diff
@@ -131,46 +157,38 @@ public static async Task<int> CreateReleaseTagsAsync(RoslynToolsSettings setting
...
+            return branchName;
         }
         catch
         {
-            return default;
+            return null;
         }
     }
 
     private static async Task<string?> TryGetRoslynCommitShaFromBuildAsync(
```
jasonmalinowski, line 186:
```suggestion
    private static async Task<string?> TryGetCommitShaFromBuildAsync(
```

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
```diff
@@ -98,30 +99,55 @@ public static async Task<int> CreateReleaseTagsAsync(RoslynToolsSettings setting
...
+
+            // First we try to get the info we want from the build directly, if we can find it
+            var buildNumber = VisualStudioRepository.GetBuildNumberFromUrl(url);
+            var pipelineName = product.GetBuildPipelineName(connection.BuildProjectName);
+            if (pipelineName is not null)
+            {
+                var builds = await connection.TryGetBuildsAsync(pipelineName, buildNumber);
+                if (builds is not null)
+                {
+                    foreach (var build in builds)
```
jasonmalinowski, line 126: When can there be more than one here? And the first one w/ a non-empty source branch is just always correct?

src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
davidwengier: I would expect it to work for anything we have a `Product` class for, as thats kinda the whole point of the abstraction. I probably copied this message from `vsbranchinfo`, and that was probably written before F# and TypeScript were added.

src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
davidwengier: Hey, I just modifed this one... didn't you originally write this? 😛

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
davidwengier: oh, this was actually just for my debugging.. it was annoying waiting for it to spin through all of the versions that weren't going to be relevant, to find one that was so I could actually see my code working :)

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
davidwengier: I don't understand how or why, but yes, in the past there have been multiple builds on the same pipeline with the same build number.

src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
davidwengier: `GetValueForOption` is annotated as `MayBeNull`, but we have an allow list and a default for the option, so it won't be. All of the commands do this.

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
davidwengier: It could, but the command handlers in general have no logic, so this makes more sense to me. It would be a literal move for no benefit IMO

src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
jasonmalinowski: I don't think so, and I'm not just saying that to get out of the blame!

src/dotnet-roslyn-tools/Commands/CreateReleaseTagsCommand.cs
```diff
@@ -13,10 +14,16 @@ internal class CreateReleaseTagsCommand
...
     private static readonly CreateReleaseTagsCommandDefaultHandler s_defaultHandler = new();
 
+    // Filter the product to only those with git credentials, as we need to be able to commit to the repo to add tags
+    private static readonly string[] s_allProductNames = VSBranchInfo.AllProducts.Where(p => p.GitUserName.Length > 0).Select(p => p.Name.ToLower()).ToArray();
+
+    internal static readonly Option<string> ProductOption = new Option<string>(new[] { "--product", "-p" }, () => "roslyn", "Which product to get info for").FromAmong(s_allProductNames);
+
     public static Symbol GetCommand()
     {
         var command = new Command("create-release-tags", "Generates git tags for VS releases in the Roslyn repo.")
```
jasonmalinowski, line 24:
```suggestion
        var command = new Command("create-release-tags", "Generates git tags for VS releases in the repo.")
```

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
jasonmalinowski: To me this is the prefect logic that should go in the command handler: it's parsing the input into the workable input. The other two parameters are already rich types, after all.

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
jasonmalinowski: That seems terrifying. Should we just throw in that case and if/when that ever happens again we can then figure out what the right action is? My concern is that might tag the wrong commit in that case and we won't actually know if this did the right thing.

jasonmalinowski (approved):

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
davidwengier: Dug up some context on this (surprisingly, Teams search actually worked!): https://teams.microsoft.com/l/message/19:ed7a508bf00c4b088a7760359f0d0308@thread.skype/1621985340130?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=4ba7372f-2799-4677-89f0-7a1aaea3706c&parentMessageId=1621985340130&teamName=.NET%20Developer%20Experience&channelName=InfraSwat&createdTime=1621985340130

So this was a bug, and was probably a one-off issue that it happened, but given that a) obviously AzDo doesn't care if you tell it something has the same build number as something else and b) this tool looks through historical builds so the bugged build could still be in play, I would be hesitant to remove this. When it did come up (it was in running the `vsbranchinfo` tool to do a VS Mac insertion) the commit that it found was correct, but I don't remember much else about it.

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
davidwengier: You're not wrong. Whats _really_ annoying is that System.CommandLine is already doing the same parsing, and validating that the input is correct and in the expected set. I'll see if there is a better API for getting strongly typed arguments out of it

src/dotnet-roslyn-tools/CreateReleaseTags/CreateReleaseTags.cs
jasonmalinowski: Even if we take for granted that we need to support this case, there's still no argument here for why this is picking the right one versus the wrong one. I'd also imagine that doesn't matter anymore since we wouldn't be trying to create release tags for either of those builds unless we actually shipped one. And we could easily resolve the problem by just deleting one of the historical builds that we know wasn't shipped.

davidwengier:
Merging this so it doesn't get stale, and so I can run it for razor for the 17.5 p3 release, but the have faith the follow up is still in my todo list.

PR #1326, opened 2023-03-03
Use WPF for ProjectDependencies

ryzngard:
Maui is cool, but was depending on an external package to add a treeview. Since we don't allow NuGet, just use WPF which has a built in TreeView


dibarbet (approved):
I won't pretend to understand all the WPF bits, but this worked locally for me first try.

PR #1384, opened 2024-02-07
Change the issue tagger to a polling model

Cosifne:
Context: 
Issue tagger is broken since Allison left.
It is using her PAT, which is not great.
So we want to switch to Dnceng so that they handle the PAT for us.

However, this brings another problem. The issue tagger is triggered by VS pipeline under DevDiv, but Azure pipeline doesn't support a cross-org trigger.

So it has to be changed into a polling model.
now the logic becomes:
Every 12 hours:
1. We fetch 20 builds from VS. (I feel it should be enough to cover every build)
2. Compare and generate the diff.
3. Create an issue.
4. Stop if we see the issue has already been created. (It means the issue is created by the previous run of the pipeline) Or any error happens.

I have run it locally.. but seems like Roslyn/Razor/TS/F# are all not inserted in the last 20 build, so nothing gets changed.
And I can't test this using my PAT in cloud. (😅I have been warned)  So I would say I prefer to check this in first, and then setup the yml with dnceng, later it can use their PAT to test

Updates: 
1. Tested locally, works https://github.com/dotnet/roslyn/issues/72198
2. I added the `--build` command back, it allows the tagger to only tag that VS build.  So if someone needs to tag a certain old build it should work.

### DIFF
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -1,0 +1,6 @@
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- script: echo Hello, world!
+  displayName: 'Run a one-line script'
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -1,3 +1,14 @@
+parameters:
+  - name: GitHubToken
+    type: string
+    default: ""
+  - name: DevDivToken
+    type: string
+    default: ""
+  - name: DncengToken
+    type: string
+    default: ""
+
 pool:
   vmImage: 'ubuntu-latest'
 
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -22,4 +22,4 @@ stages:
       - task: UseDotNet@2
       - script: dotnet tool install Microsoft.RoslynTools --prerelease --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json -g
         displayName: Install Roslyn Insertion Tool
-      - script: echo "Hello world"
+      - script: roslyn-tools pr-tagger -b main-34602.64 -c 76ef32b4c79be68ee58633b9ff25e28be110ff43 --github-token $(GitHubToken) --devdiv-azdo-token $(DevDivToken) --dnceng-azdo-token $(DncengToken)
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -1,3 +1,14 @@
+resources:
+  pipelines:
+    - pipeline: vs-pipeline
+      source: DD-Official\Release\DD-CB-TestSignVS
+      project: DevDiv
+      trigger:
+        branches:
+          include:
+            - main
+            - rel/*
+
 parameters:
   - name: GitHubToken
     type: string
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -1,24 +1,10 @@
-resources:
-  pipelines:
-    - pipeline: vs-pipeline
-      source: DD-Official\Release\DD-CB-TestSignVS
-      project: DevDiv
-      trigger:
-        branches:
-          include:
-            - main
-            - rel/*
-
-parameters:
-  - name: GitHubToken
-    type: string
-    default: ""
-  - name: DevDivToken
-    type: string
-    default: ""
-  - name: DncengToken
-    type: string
-    default: ""
+schedules:
+- cron: "0 */12 * * *"
+  displayName: Roslyn Tagger Tool
+  branches:
+    include:
+    - main
+  always: true
 
 pool:
   vmImage: 'ubuntu-latest'
@@ -33,5 +19,5 @@ stages:
       - task: UseDotNet@2
       - script: dotnet tool install Microsoft.RoslynTools --prerelease --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json -g
         displayName: Install Roslyn Insertion Tool
-      - script: roslyn-tools pr-tagger -b main-34526.79 -c 52f1120f305584c4b4e72a4948d6a37077d97d94 --github-token ${{ parameters.GitHubToken }} --devdiv-azdo-token ${{ parameters.DevDivToken }} --dnceng-azdo-token ${{ parameters.DncengToken }}
+      - script: # roslyn-tools pr-tagger --github-token "" --devdiv-azdo-token "" --dnceng-azdo-token ""
         displayName: Run PR Tagger
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -3,7 +3,8 @@ pool:
 
 stages:
 - stage: tagInsertedPullRequest
-  pool: ubuntu-latest
+  pool:
+    vmImage: 'ubuntu-latest'
   jobs:
   - job: runPRTagger
     steps:
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -23,3 +23,4 @@ stages:
       - script: dotnet tool install Microsoft.RoslynTools --prerelease --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json -g
         displayName: Install Roslyn Insertion Tool
       - script: roslyn-tools pr-tagger -b main-34602.64 -c 76ef32b4c79be68ee58633b9ff25e28be110ff43 --github-token ${{ parameters.GitHubToken }} --devdiv-azdo-token ${{ parameters.DevDivToken }} --dnceng-azdo-token ${{ parameters.DncengToken }}
+        displayName: Run PR Tagger
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -1,6 +1,13 @@
 pool:
   vmImage: 'ubuntu-latest'
 
-steps:
-- script: echo Hello, world!
-  displayName: 'Run a one-line script'
+stages:
+- stage: Tag inserted pull request
+  pool: ubuntu-latest
+  jobs:
+  - job: RunPRTagger
+    steps:
+      - task: UseDotNet@2
+      - script: dotnet tool install Microsoft.RoslynTools --prerelease --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json -g
+        displayName: Install Roslyn Insertion Tool
+      - script: echo "Hello world"
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -2,10 +2,10 @@ pool:
   vmImage: 'ubuntu-latest'
 
 stages:
-- stage: Tag inserted pull request
+- stage: tagInsertedPullRequest
   pool: ubuntu-latest
   jobs:
-  - job: RunPRTagger
+  - job: runPRTagger
     steps:
       - task: UseDotNet@2
       - script: dotnet tool install Microsoft.RoslynTools --prerelease --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json -g
--- a/insertion-pr-tagger.yml
+++ b/insertion-pr-tagger.yml
@@ -22,5 +22,5 @@ stages:
       - task: UseDotNet@2
       - script: dotnet tool install Microsoft.RoslynTools --prerelease --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json -g
         displayName: Install Roslyn Insertion Tool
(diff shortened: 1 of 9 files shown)


insertion-pr-tagger.yml
```diff
@@ -0,0 +1,23 @@
...
+- stage: tagInsertedPullRequest
+  pool:
+    vmImage: 'ubuntu-latest'
+  jobs:
+  - job: runPRTagger
+    steps:
+      - task: UseDotNet@2
+      - script: dotnet tool install Microsoft.RoslynTools --prerelease --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json -g
+        displayName: Install Roslyn Insertion Tool
+      - script: # roslyn-tools pr-tagger --github-token "" --devdiv-azdo-token "" --dnceng-azdo-token ""
```
Cosifne, line 22: Need to set the `PAT` after this get merged with dnceng, we are going to use bot's PAT.

Cosifne:
Seems Octokit is not allowed ☹️ I will switch to http client

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -2,133 +2,186 @@
...
+    public static async Task<ImmutableArray<(string vsBuild, string vsCommit, string previousVsCommitSha)>> GetVSBuildsAndCommitsAsync(
+        AzDOConnection devdivConnection,
+        ILogger logger,
+        int vsBuildNumber,
+        CancellationToken cancellationToken)
+    {
+        var builds = await devdivConnection.TryGetBuildsAsync(
+            "DD-CB-TestSignVS",
+            logger: logger,
+            maxBuildNumberFetch: vsBuildNumber,
```
dibarbet, line 163: IMHO 20 might not be enough, especially if we're only running every 12 hours.  And we'll just silently miss builds if it isn't enough.
Is it possible to instead fetch all builds up until the last time we ran (e.g. fetch all builds from the last 12 hours)?

Cosifne: I feel each build would takes 1 to 1.5 hour. So if we run it every 12 hours, would it be enough?
It's possible to keep find the last time it gets run, but I am littble bit worry about if
for example: the last insertion issue is created maybe a few months ago and the tagger stops working somehow, later it is fixed, then it will start to generate hundreds of issue when it starts to run

Cosifne: OK, how about this, we use both
1. A max number threshold. (Can be specify from the command line)
2. Keep finding the last time we ran. Stop if it is found or the max number is hit.

In this case I feels we should be good?

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -188,40 +241,54 @@ private static async Task<GitRepository> GetVSRepositoryAsync(GitHttpClient gitC
...
+    private static async Task<bool> HasIssueAlreadyCreatedAsync(
+        HttpClient client,
+        string repoName,
+        string title,
+        ILogger logger)
+    {
+        var response = await client.GetAsync($"search/issues?q={title}+repo:dotnet/{repoName}+is:issue+label:{InsertionLabel}").ConfigureAwait(false);
+        if (!response.IsSuccessStatusCode)
+        {
+            logger.LogError("Failed to search on GitHub");
```
JoeRobich, line 283: Would be good to include the status code here.

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -2,133 +2,186 @@
...
+            "DD-CB-TestSignVS",
+            logger: logger,
+            maxBuildNumberFetch: vsBuildNumber,
+            resultsFilter: BuildResult.Succeeded,
+            buildQueryOrder: BuildQueryOrder.FinishTimeDescending).ConfigureAwait(false);
+        var vsRepository = await GetVSRepositoryAsync(devdivConnection.GitClient);
+        if (builds is not null)
+        {
+            // Find previous VS commit SHA
+            var buildInfoTask = builds.Select(async build =>
```
JoeRobich, line 170: Do we need to check if the build is for the main branch?

src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
```diff
@@ -49,12 +44,30 @@ public async Task<int> InvokeAsync(InvocationContext context)
...
             return await PRTagger.PRTagger.TagPRs(
-                vsBuild,
-                vsCommitSha,
+                vsBuildsAndCommitSha: buildsAndCommits,
                 settings,
-                logger,
-                CancellationToken.None).ConfigureAwait(false);
+                devdivConnection,
+                client,
+                logger).ConfigureAwait(false);
```
JoeRobich, line 70: I would move this bit to a new method in the PRTagger class and call into it here. The commands are intended to be light on logic other than option parsing.

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -2,133 +2,186 @@
...
+        // 3. If we found the issue with the same title has been created. It means the issue is created because the last run of the tagger.
         foreach (var product in VSBranchInfo.AllProducts)
         {
-            // We currently only support creating issues for GitHub repos
-            if (!product.RepoHttpBaseUrl.Contains("github.com"))
+            foreach (var (vsBuild, vsCommitSha, previousVsCommitSha) in vsBuildsAndCommitSha)
             {
-                continue;
+                var result = await TagProductAsync(product, logger, vsCommitSha, vsBuild, previousVsCommitSha, settings, devdivConnection, dncengConnection, gitHubClient).ConfigureAwait(false);
+                if (result is TagResult.Failed or TagResult.IssueAlreadyCreated)
```
JoeRobich, line 52: Is there a goal to tag builds that were missed while the tagger was down? On first run we will tag the first n builds, but then we can never go back further because on subsequent runs we will hit `IssueAlreadyCreated` results that stop us from going back further.

src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
```diff
@@ -2,29 +2,27 @@
...
     {
         var command = new Command("pr-tagger", "Tags PRs inserted in a given VS build.")
         {
-            VSBuild,
-            CommitId,
             CommonOptions.GitHubTokenOption,
             CommonOptions.DevDivAzDOTokenOption,
-            CommonOptions.DncEngAzDOTokenOption
+            CommonOptions.DncEngAzDOTokenOption,
+            maxVsBuildCheckNumber
```
JoeRobich, line 25: I believe this ordering influences how the help (-?) text is generated. You may want the command specific option to show before the general options.

src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
```diff
@@ -2,29 +2,27 @@
...
+
+namespace Microsoft.RoslynTools.Commands;
 
 internal static class PRTaggerCommand
 {
     private static readonly PRTaggerCommandDefaultHandler s_prTaggerCommandHandler = new();
-
-    private static readonly Option<string> VSBuild = new(new[] { "--build", "-b" }, "VS build number") { IsRequired = true };
-    private static readonly Option<string> CommitId = new(new[] { "--commit", "-c" }, "VS build commit") { IsRequired = true };
+    public static readonly Option<int> maxVsBuildCheckNumber = new(new[] { "--vsBuildCheckNumber" }, () => 20, "Maximum number of VS build to check. Tagger would compare each VS build and its parent commit to find the inserted payload.");
```
JoeRobich, line 16: What is the most builds that they've had in a day? Will 40 be enough?

JoeRobich (commented):
Just some thoughts.

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
Cosifne: I don't think so, if it merged into other branch, it should still work.

src/dotnet-roslyn-tools/Commands/PRTaggerCommand.cs
Cosifne: It generated about 30 per day. but I will increase this number to 50 to make sure things are correct

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
JoeRobich: Shame we can't limit our build query to those requestedBy DotNet Bot. But I guess we could miss joint PRs that were created by other teams.

arkalyanms (approved):

arkalyanms:
if partially working, consider merging and iteratively improving

Cosifne:
> if partially working, consider merging and iteratively improving

I have some thoughts on improving this, will back to work on this after I moved the format repo

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
Cosifne: So one update on this:
AzDo allows searching for a build, or getting the latest `n` builds.
So there are two ways:
1. Fetch the max number of build, then search the build we need in the list.
2. Find the build by the build number, then compare the how many builds there are since the build has completed.

I use `1` here because I feel it feels more natural to fit the goal 
```
A max number threshold. (Can be specify from the command line)
Keep finding the last time we ran. Stop if it is found or the max number is hit.
```

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
Cosifne: I added another `--build`  command, allow you can create based on build. (Like the old tagger does)

Cosifne:
Ping @dibarbet @JoeRobich  to review : )

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -2,133 +2,256 @@
...
+        // 1. All the VS build has been checked.
+        // 2. If some errors happens.
+        // 3. If we found the issue with the same title has been created. It means the issue is created because the last run of the tagger.
         foreach (var product in VSBranchInfo.AllProducts)
         {
             // We currently only support creating issues for GitHub repos
             if (!product.RepoHttpBaseUrl.Contains("github.com"))
             {
-                continue;
+                return -1;
```
JoeRobich, line 57: This change would stop us from tagging later Products that may live in GitHub. Could log a warning and continue to the next Product.

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -2,133 +2,256 @@
...
+        var prDescription = new StringBuilder();
+        var isSuccess = PRFinder.PRFinder.FindPRs(previousProductCommitSha, currentProductCommitSha, PRFinder.PRFinder.DefaultFormat, logger, gitHubRepoPath, prDescription);
+        if (isSuccess != 0)
+        {
+            // Error occurred; should be logged in FindPRs method
+            return TagResult.Failed;
+        }
+
+        var issueTitle = $"[Automated] PRs inserted in VS build {vsBuild}";
+        var hasIssueAlreadyCreated = await HasIssueAlreadyCreatedAsync(remoteConnections.GitHubClient, gitHubRepoName, issueTitle, logger).ConfigureAwait(false);
```
JoeRobich, line 157: Should this check happen earlier to avoid making any requests that aren't necessary?

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -188,40 +311,125 @@ private static async Task<GitRepository> GetVSRepositoryAsync(GitHttpClient gitC
...
+    /// <summary>
+    /// Check if the issue with <param name="title"/> exists in repo.
+    /// </summary>
+    private static async Task<bool> HasIssueAlreadyCreatedAsync(
+        HttpClient client,
+        string repoName,
+        string title,
+        ILogger logger)
+    {
+        var response = await client.GetAsync($"search/issues?q={title}+repo:dotnet/{repoName}+is:issue+label:{InsertionLabel}").ConfigureAwait(false);
```
JoeRobich, line 350: Can this just call into `SearchIssuesOnGitHubAsync` so that there is only one bit of code responsible for searching issues?

JoeRobich (commented):
I think it would be fine to move forward with this, but, depending on how complete we want to make things, do encourage brainstorming on how we can backfill insertions we've missed.

src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -188,40 +311,125 @@ private static async Task<GitRepository> GetVSRepositoryAsync(GitHttpClient gitC
...
+            throw new Exception($"Error happens when try to search {title} in {repoName}.");
+        }
+
+        var content = await response.Content.ReadAsStringAsync();
+        var jsonResponseContent = JsonObject.Parse(content)!;
+        var issueNumber = int.Parse(jsonResponseContent["total_count"]!.ToString());
+        return issueNumber != 0;
+    }
+
+    private static async Task<string?> FindTheLastReportedVSBuildAsync(
```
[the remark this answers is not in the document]
Cosifne: I will leave this comment open. If we hit problem in this current impl we can try this way

PR #1390, opened 2024-03-15
Fix roslyntool

genlu:


NuGet.config
```diff
@@ -4,13 +4,19 @@
     <clear />
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
-    <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
-    <add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
+    <add key="dotnet6" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json" />
```
genlu, line 7: I got a bunch of missing package error when build locally, so I copied the feeds from Roslyn

src/dotnet-roslyn-tools/Products/Roslyn.cs
```diff
@@ -23,7 +23,7 @@ internal class Roslyn : IProduct
         => buildProjectName switch
         {
             "DevDiv" => "Roslyn-Signed",
-            "internal" => "dotnet-roslyn CI",  // dnceng
+            "internal" => "dotnet-roslyn-official",  // dnceng
```
genlu, line 26: this is the main issue

JoeRobich (approved):

NuGet.config
JoeRobich: We are going to find out System.CommandLine.Experimental 0.1.0-alpha-63227-01 was in the dotnet5 feed that you removed.

arkalyanms (approved):

NuGet.config
genlu: Yes indeed 😅

PR #1407, opened 2024-04-10
Gracefully handle absence of .corext/Configs/default.config

RikkiGibson:
This is in preparation for an upcoming VS change which is removing certain corext files from the repo. As part of that work we expect to also start using Microsoft.Net.Compilers.Toolset for the toolset instead of VS.Tools.Roslyn.

The particular Roslyn change which will copy Microsoft.Net.Compilers.Toolset to VSSetup/DevDivPackages in the official build I haven't gotten working yet. However I think we might still be able to use RIT to insert successfully without it. We can't adjust the specific code paths for updating the toolset until we get the new toolset shipping in that directory though.

Test inserting into branch with no default.config: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/543224
Compare with insertion into main: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/543205?_a=files

Testing vs build with new toolset package: https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/543201

@dibarbet @Cosifne @jaredpar for review

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
@@ -64,10 +64,12 @@ namespace Roslyn.Insertion
                     download: true,
                     versionDescriptor: vsBranch);
                 defaultConfigOriginal = await new StreamReader(defaultConfigStream).ReadToEndAsync();
+                Console.WriteLine($"'{DefaultConfigPath}' found.");
             }
             catch (VssServiceException ex) when (ex.IsFileNotFound())
             {
                 defaultConfigOriginal = null;
+                Console.WriteLine($"'{DefaultConfigPath}' not found. Will search package props files for current versions instead.");
             }
 
             ComponentToFileMap = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
--- a/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
@@ -11,6 +11,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using System.Xml.Linq;
 using Microsoft.TeamFoundation.SourceControl.WebApi;
+using Microsoft.VisualStudio.Services.Common;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using NuGet.Versioning;
@@ -39,14 +40,14 @@ namespace Roslyn.Insertion
         private const string PackagePropsPath = "Directory.Packages.props";
         private const string LegacyPackagePropsPath = "Packages.props";
 
-        private readonly string _defaultConfigOriginal;
+        private readonly string? _defaultConfigOriginal;
 
-        public XDocument ConfigDocument { get; }
+        public XDocument? ConfigDocument { get; }
 
-        private CoreXT(string configOriginalText)
+        private CoreXT(string? configOriginalText)
         {
             _defaultConfigOriginal = configOriginalText;
-            ConfigDocument = XDocument.Parse(configOriginalText, LoadOptions.None);
+            ConfigDocument = configOriginalText is null ? null : XDocument.Parse(configOriginalText, LoadOptions.None);
         }
 
         public static async Task<CoreXT> Load(GitHttpClient gitClient, string commitId)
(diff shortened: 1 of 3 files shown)


RikkiGibson:
@Cosifne @dibarbet Please take a look as VS plans to merge the change to remove default.config soon

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -39,27 +40,35 @@ internal class CoreXT
...
-            using var defaultConfigStream = await gitClient.GetItemContentAsync(
-                vsRepoId,
-                DefaultConfigPath,
-                download: true,
-                versionDescriptor: vsBranch);
-            var defaultConfigOriginal = await new StreamReader(defaultConfigStream).ReadToEndAsync();
+            string? defaultConfigOriginal;
+            try
+            {
+                using var defaultConfigStream = await gitClient.GetItemContentAsync(
```
dibarbet, line 61: if there is no default.config, where are versions for things like VS.ExternalAPIs.Roslyn stored?    Does it just come from roslyn.props?

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -39,27 +40,35 @@ internal class CoreXT
...
+                    vsRepoId,
+                    DefaultConfigPath,
+                    download: true,
+                    versionDescriptor: vsBranch);
+                defaultConfigOriginal = await new StreamReader(defaultConfigStream).ReadToEndAsync();
+            }
+            catch (VssServiceException ex) when (ex.IsFileNotFound())
+            {
+                defaultConfigOriginal = null;
```
dibarbet, line 70: consider logging something when there is no default.config (I assume this is expected for main, but unexpected for servicing brnaches)?

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
```diff
@@ -39,27 +40,35 @@ internal class CoreXT
...
 
-        public XDocument ConfigDocument { get; }
+        public XDocument? ConfigDocument { get; }
 
-        private CoreXT(string configOriginalText)
+        private CoreXT(string? configOriginalText)
         {
             _defaultConfigOriginal = configOriginalText;
-            ConfigDocument = XDocument.Parse(configOriginalText, LoadOptions.None);
+            ConfigDocument = configOriginalText is null ? null : XDocument.Parse(configOriginalText, LoadOptions.None);
```
dibarbet, line 50: Was wondering if this should be something that gets passed in as a parameter to the tool.  We'd update our yaml in main to pass it in (once VS switches over).  New servicing branches would automatically get the value when main merges in, but old ones would continue to use the old behavior

then we could throw on unexpected behavior (e.g. branch was supposed to have a default.config file but we didn't getone)

dibarbet (approved):

Cosifne:
🤔Just to check, the future plan is Microsoft.CodeAnalysis.Compilers would carry the complier payload (Microsoft.Net.Compilers.Toolset) into VS? (So it means we always update the toolset version in every insertion?)

Cosifne (approved):
One minor question about the future plan. Change itself looks good. : )

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
RikkiGibson: Yes, `src/ConfigData/Packages/roslyn.props` becomes the source of truth for this.

If we know of any other components still using this tool, I should perhaps test with them too.

RikkiGibson:
> 🤔Just to check, the future plan is Microsoft.CodeAnalysis.Compilers would carry the complier payload (Microsoft.Net.Compilers.Toolset) into VS? (So it means we always update the toolset version in every insertion?)

Yes though it appears that VS has been building without corext in main since at least sometime in 2023, so we've been using the compiler included in msbuild to compile without knowing. I'm working on adjusting the VS build to use Microsoft.Net.Compilers.Toolset.

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
RikkiGibson: I will punt on this for now. Thanks

src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs
RikkiGibson: Done

PR #1414, opened 2024-04-23
Fix RIT

Cosifne:
So RIT has some problem, example run: https://dev.azure.com/dnceng/internal/_apps/hub/ms.vss-releaseManagement-web.cd-release-progress?_a=release-pipeline-progress&releaseId=42298

1. After we move to 1ES template, artifacts container would become pipeline artifacts.
2. There is a problem to get pipeline artifacts directly by calling `build.GetArtifactContentZipAsync()`

I locally verified it works. 
https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/546136

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -361,7 +361,7 @@ namespace Roslyn.Insertion
                 {
                     // artifact.Resource.Data should be available and non-null due to BuildWithValidArtifactsAsync,
                     // which checks this precondition
-                    if (!StringComparer.OrdinalIgnoreCase.Equals(artifact.Resource.Type, "container"))
+                    if (!StringComparer.OrdinalIgnoreCase.Equals(artifact.Resource.Type, "pipelineArtifact"))
                     {
                         throw new InvalidOperationException($"Could not find artifact '{arcadeArtifactName}' associated with build '{build.Id}'");
                     }
--- a/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
+++ b/src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
@@ -8,7 +8,9 @@ using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
@@ -344,7 +346,6 @@ namespace Roslyn.Insertion
             }
 
             var buildClient = ComponentBuildConnection.GetClient<BuildHttpClient>();
-
             Debug.Assert(ReferenceEquals(build,
                 (await GetInsertableComponentBuildsAsync(buildClient, cancellationToken, new[] { build })).Single()));
 
@@ -359,9 +360,11 @@ namespace Roslyn.Insertion
             {
                 if (artifact.Name == arcadeArtifactName)
                 {
+                    var artifactType = artifact.Resource.Type;
                     // artifact.Resource.Data should be available and non-null due to BuildWithValidArtifactsAsync,
                     // which checks this precondition
-                    if (!StringComparer.OrdinalIgnoreCase.Equals(artifact.Resource.Type, "pipelineArtifact"))
+                    if (!artifactType.Equals("container", StringComparison.OrdinalIgnoreCase) &&
+                        !artifactType.Equals("pipelineArtifact", StringComparison.OrdinalIgnoreCase))
                     {
                         throw new InvalidOperationException($"Could not find artifact '{arcadeArtifactName}' associated with build '{build.Id}'");
                     }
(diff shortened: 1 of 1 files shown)


src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -359,9 +359,11 @@ internal static async Task<InsertionArtifacts> GetInsertionArtifactsAsync(Build
             {
                 if (artifact.Name == arcadeArtifactName)
                 {
+                    var artifactType = artifact.Resource.Type;
                     // artifact.Resource.Data should be available and non-null due to BuildWithValidArtifactsAsync,
                     // which checks this precondition
-                    if (!StringComparer.OrdinalIgnoreCase.Equals(artifact.Resource.Type, "container"))
+                    if (!artifactType.Equals("container", StringComparison.OrdinalIgnoreCase) &&
+                        !artifactType.Equals("pipelineArtifact", StringComparison.OrdinalIgnoreCase))
```
Cosifne, line 366: I guess this matches this field https://github.com/dotnet/roslyn/blob/cc6e5bd358cbf9fccde8f017446d9843744f058f/azure-pipelines-official.yml#L164

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -395,21 +397,45 @@ private static async Task<string> DownloadBuildArtifactsAsync(BuildHttpClient bu
...
-            using (var ms = new MemoryStream())
+            Stopwatch watch = Stopwatch.StartNew();
+            if (artifactType.Equals("container", StringComparison.OrdinalIgnoreCase))
             {
-                await s.CopyToAsync(ms);
-                using (ZipArchive archive = new ZipArchive(ms))
-                {
-                    archive.ExtractToDirectory(tempDirectory);
-                }
+                using var contentStream = await buildClient.GetArtifactContentZipAsync(Options.ComponentBuildProjectNameOrFallback, build.Id, artifact.Name, cancellationToken);
```
Cosifne, line 406: This is just the original logic.

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
dibarbet: wondering - why are our insertions not currently erroring?

Cosifne: I belive if you run RIT from the signed build pipeline, it find the VSSetup folder locally.
But when running it offline, it will try to download it

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
```diff
@@ -395,21 +397,45 @@ private static async Task<string> DownloadBuildArtifactsAsync(BuildHttpClient bu
...
+                await ExtractToDirectoryAsync(contentStream, tempDirectory).ConfigureAwait(false);
+                Console.WriteLine($"Artifact download took {watch.ElapsedMilliseconds / 1000} seconds");
+                return archiveDownloadPath;
+            }
+            else
+            {
+                // When the published by Publish artifacts pipeline task, buildClient.GetArtifactContentZipAsync() is unable to get the content of the artifacts.
```
dibarbet, line 413: Interesting that we download the artifact directly.  It would be nice if instead we could utilize the azdo pipeline download steps before running the tool, and giving the tool a path to the predownloaded artifact.

I don't know if that would break a bunch of things though.

Cosifne: 😄Probably the tool was originally created when Azp doesn't support yml? (Just too old)

Cosifne: I only know we use it here https://dev.azure.com/dnceng/internal/_release?definitionId=102&view=mine&_a=releases but depends on the other teams's usage, if we move this to yml we might be pinged many times by other teams

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
RikkiGibson: I think I experienced this same issue when running locally, which I dealt with by downloading the artifact manually and passing the path to it.

Cosifne: So should we consider the change here is OK now? (it's not a blocking issue for our daily insertion)

RikkiGibson: I think we should take your PR, I'll wrap up reviewing it shortly

src/RoslynInsertionTool/RoslynInsertionTool/RoslynInsertionTool.VisualStudioTeamServices.cs
RikkiGibson: It looks like copy to MemoryStream was removed, was this just felt to be an unnecessary step? (and I guess one that might consume a lot of memory..)

PR #1418, opened 2024-05-06
Fix tagger

Cosifne:


src/dotnet-roslyn-tools/PRTagger/PRTagger.cs
```diff
@@ -420,6 +420,7 @@ private static int TotalCountNumber(JsonNode response)
         var response = await client.GetAsync(query).ConfigureAwait(false);
         var content = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
         var jsonResponseContent = JsonObject.Parse(content)!;
+        logger.LogInformation($"Response object is {jsonResponseContent.ToJsonString()}.");
```
Cosifne, line 423: https://dev.azure.com/dnceng/internal/_build/results?buildId=2445617&view=logs&j=9d82d844-45a0-5d2e-bd4b-992528a7bed5&t=5cca82b3-8182-5619-1c3b-68ef1449fa53
I can't repro this error locally using my token. From the error, I feel it means Github doesn't return anything. It is very likely caused by the rate limit.

So add more logging here to see what happens.

dibarbet (approved):

PR #1423, opened 2024-05-29
Exclude libTo2Git.so

Cosifne:
Exclude the symbol publish for libTo2Git.so, following the instructions [here](https://github.com/dotnet/arcade/blob/main/Documentation/CorePackages/Publishing.md#can-we-exclude-symbols-from-publishing-to-symbols-server)
Fix https://github.com/dotnet/dnceng/issues/2958#issuecomment-2137692339


Cosifne:
Get test run passed. https://dev.azure.com/dnceng/internal/_build/results?buildId=2462758&view=results

PR #1431, opened 2024-06-26
Don't update XAML version

genlu:
Issuer is fixed in [this PR](https://devdiv.visualstudio.com/DevDiv/_git/a290117c-5a8a-40f7-bc2c-f14dbe3acf6d/pullrequest/558810)

### DIFF
--- a/src/RoslynInsertionTool/RoslynInsertionTool.Commandline/App.config
+++ b/src/RoslynInsertionTool/RoslynInsertionTool.Commandline/App.config
@@ -94,7 +94,7 @@
         <value>False</value>
       </setting>
       <setting name="UpdateXamlRoslynVersion" serializeAs="String">
-        <value>True</value>
+        <value>False</value>
       </setting>
     </RoslynInsertionTool.Settings>
   </applicationSettings>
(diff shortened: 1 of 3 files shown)


arkalyanms (approved):

src/RoslynInsertionTool/RoslynInsertionTool.Commandline/App.config
```diff
@@ -94,7 +94,7 @@
         <value>False</value>
       </setting>
       <setting name="UpdateXamlRoslynVersion" serializeAs="String">
-        <value>True</value>
+        <value>False</value>
```
jjonescz, line 97: Should we update also `src/RoslynInsertionTool/RoslynInsertionTool.Commandline/Settings.settings`?

src/RoslynInsertionTool/RoslynInsertionTool.Commandline/App.config
jjonescz, line 1: > Issuer is fixed in [this PR](https://devdiv.visualstudio.com/DevDiv/_git/a290117c-5a8a-40f7-bc2c-f14dbe3acf6d/pullrequest/558810)

Are we sure it's fixed? It doesn't look like they tried inserting with the version not matching. But I guess we will see and can always change the default back.

jjonescz (approved):

src/RoslynInsertionTool/RoslynInsertionTool.Commandline/App.config
genlu: Yea, I tried to test their fix in this PR, which still failed with build error. I ask them to take a look
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_git/VS/pullrequest/560567

genlu: Tested again, now it seems to be fixed
https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/563574

PR #1443, opened 2024-10-04
Remove merge from 17.6 to 17.8 and 17.8 to 17.10

akhera99:
Removing to prevent reverting

### DIFF
--- a/src/GitHubCreateMergePRs/config.xml
+++ b/src/GitHubCreateMergePRs/config.xml
@@ -8,8 +8,6 @@
     <!-- Roslyn servicing branches (flowing between releases) -->
     <merge from="dev15.9.x" to="release/dev16.11-vs-deps" />
     <merge from="release/dev16.11-vs-deps" to="release/dev17.6" />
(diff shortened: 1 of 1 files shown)


src/GitHubCreateMergePRs/config.xml
```diff
@@ -8,8 +8,6 @@
     <!-- Roslyn servicing branches (flowing between releases) -->
     <merge from="dev15.9.x" to="release/dev16.11-vs-deps" />
     <merge from="release/dev16.11-vs-deps" to="release/dev17.6" />
-    <merge from="release/dev17.6" to="release/dev17.8" />
-    <merge from="release/dev17.8" to="release/dev17.10" />
```
333fred, line 12: Did we want to kill the 17.8->17.10 flow as well?

akhera99: We do for now at least since we don't want this PR to flow forward https://github.com/dotnet/roslyn/pull/75348.

333fred (approved):

PR #1499, opened 2025-08-18
Allow create-release-tags to looks up repo SHA from vmr build

JoeRobich:
Starting with .NET 10 preview 5, the SDK has built out of the dotnet VMR. This broke release tagging for the SDK. The SDK release tagger is updated to look up the repo specific SHA from the VMR build commit. Tags from VMR builds call out specifically that the information is for the VMR build ([example](https://github.com/dotnet/roslyn/releases/tag/NET-SDK-10.0.100-preview.7.25380.108)).

```
NET-SDK-10.0.100-preview.7.25380.108: Roslyn VMR Version: 5.0.0-2.25373.105
VMR Commit SHA: b64f1193459ddeba61d78aa61f90604410734e0f
VMR Internal ID: 20250723.105
SDK Tag: v10.0.100-preview.7.25380.108
```

### DIFF
--- a/src/dotnet-roslyn-tools/CreateReleaseTags/AbstractReleaseTagger.cs
+++ b/src/dotnet-roslyn-tools/CreateReleaseTags/AbstractReleaseTagger.cs
@@ -10,11 +10,12 @@ using Microsoft.RoslynTools.Utilities;
 
 namespace Microsoft.RoslynTools.CreateReleaseTags;
 
-internal abstract class AbstractReleaseTagger<TRelease, TBuild>
+internal abstract class AbstractReleaseTagger<TRelease, TBuild>(ILogger logger)
     where TRelease : ReleaseInformation
     where TBuild : BuildInformation
 {
     public abstract string Name { get; }
+    protected ILogger Logger => logger;
     public abstract Task<ImmutableArray<TRelease>> GetReleasesAsync(RemoteConnections connections);
     public abstract string GetTagName(TRelease release);
     public abstract Task<TBuild?> TryGetBuildAsync(RemoteConnections connections, IProduct product, TRelease release);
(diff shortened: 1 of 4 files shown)


src/dotnet-roslyn-tools/CreateReleaseTags/SdkReleaseTagger.cs
```diff
@@ -112,6 +112,38 @@ static string ConvertVersionToBuildId(string version)
...
+        // This is used when the SHA returned from the Release build does not exist in the target repo. In SDK scenarios it
+        // gives us a chance to query the dotnet VMR and identify the SHA for our repo.
+        VmrSourceManifest? sourceManifest = null;
+
+        try
+        {
+            sourceManifest = await connections.GitHubClient
+                .GetFromJsonAsync<VmrSourceManifest?>($"https://raw.githubusercontent.com/dotnet/dotnet/{vmrBuild.CommitSha}/src/source-manifest.json");
+        }
+        catch (Exception) { }
```
dibarbet, line 126: Should this log anything?  For example if we get rate limited it seems like we'd never see it.

src/dotnet-roslyn-tools/CreateReleaseTags/AbstractReleaseTagger.cs
```diff
@@ -53,21 +54,53 @@ public async Task CreateReleaseTagsAsync(
...
             }
 
-            logger.LogInformation("Tagging {CommitSha} as '{TagName}'.", build.CommitSha, tagName);
+            if (TryCreateTag(release, build, tagName))
+            {
+                tagsAdded++;
+            }
+            else
+            {
+                await TryTagFromVmrBuildAsync(release, build, tagName);
```
dibarbet, line 63: Instead of trying to query both, it looks like the AzDo api exposes the build repository - https://learn.microsoft.com/en-us/rest/api/azure/devops/build/builds/get?view=azure-devops-rest-7.1#build

Could we instead check that to determine if we need to specially handle the vmr?

dibarbet: i was confused - `build` is not an azdo build, but a tag from the SDK

dibarbet (approved):

